PR #53, opened 2019-04-23
#52 log the payload received from adyen in case of error

### LINKED ISSUE #52: Log JSON payload if a JSON parsing error occured
Right now if this line fails
https://github.com/commercetools/commercetools-adyen-integration/blob/a6959d906ab6cc5050b872daefb6de40f88cb071/notification/src/api/notification/notification.controller.js#L17

we will log it like this:

```
err: {
 message:  "Unexpected number in JSON at position 652"    
 name:  "SyntaxError"    
 stack:  "SyntaxError: Unexpected number in JSON at position 652
  at JSON.parse (<anonymous>)
  at handleNotification (/app/src/api/notification/notification.controller.js:16:38)
  at process._tickCallback (internal/process/next_tick.js:68:7)"    
}
```

It's impossible to identify what was exactly wrong with the payload that it failed to parse. 
We should catch the JSON parsing separately and in this case log the payload too.

LEQADA:
Resolves #52

Not the log looks like this:

```
{
  "name": "ctp-adyen-integration-notifications",
  "hostname": "local",
  "pid": 32801,
  "level": 50,
  "adyenRequestBody": "{\r\n  \"live\": \"false\",\r\n  \"notificationItems\": [\r\n    {\r\n      \"NotificationRequestItem\": {\r\n        \"additionalData\": {\r\n          \"expiryDate\": \"12\\/2012,\r\n          \" NAME1 \": \"VALUE1\",\r\n          \"authCode\": \"1234\",\r\n          \"cardSummary\": \"7777\",\r\n          \"totalFraudScore\": \"10\",\r\n          \"NAME2\": \"  VALUE2  \",\r\n          \"fraudCheck-6-ShopperIpUsage\": \"10\"\r\n        },\r\n        \"amount\": {\r\n          \"currency\": \"EUR\",\r\n          \"value\": 10100\r\n        },\r\n        \"eventCode\": \"AUTHORISATION\",\r\n        \"eventDate\": \"2019-01-30T18:16:22+01:00\",\r\n        \"merchantAccountCode\": \"XXX\",\r\n        \"merchantReference\": interfaceIds,\r\n        \"operations\": [\r\n          \"CANCEL\",\r\n          \"CAPTURE\",\r\n          \"REFUND\"\r\n        ],\r\n        \"paymentMethod\": \"visa\",\r\n        \"pspReference\": \"psp_reference\",\r\n        \"reason\": \"1234:7777:12\\/2012\",\r\n        \"success\": \"true\"\r\n      }\r\n    }\r\n  ]\r\n}\r\n",
  "err": {
    "message": "Unexpected token \r in JSON at position 155",
    "name": "SyntaxError",
    "stack": "SyntaxError: Unexpected token \r in JSON at position 155\n    at JSON.parse (<anonymous>)\n    at handleNotification (/app/src/api/notification/notification.controller.js:17:38)\n    at process._tickCallback (internal/process/next_tick.js:178:7)"
  },
  "msg": "Unexpected exception occurred.",
  "time": "2019-04-23T16:33:57.473Z",
  "v": 0
}
```

### DIFF
--- a/notification/src/api/notification/notification.controller.js
+++ b/notification/src/api/notification/notification.controller.js
@@ -20,9 +20,9 @@ async function handleNotification (request, response) {
       200,
       { 'Content-Type': 'application/json' },
       JSON.stringify({ notificationResponse: '[accepted]' }))
-  } catch (e) {
-    logger.error(e,
-      `Unexpected exception occurred. Received payload from Adyen: ${body}`)
+  } catch (err) {
+    logger.error({ adyenRequestBody: `${body}`, err },
+      'Unexpected exception occurred.')
     return httpUtils.sendResponse(response, 500)
   }
 }
--- a/notification/src/api/notification/notification.controller.js
+++ b/notification/src/api/notification/notification.controller.js
@@ -21,7 +21,8 @@ async function handleNotification (request, response) {
       { 'Content-Type': 'application/json' },
       JSON.stringify({ notificationResponse: '[accepted]' }))
   } catch (e) {
-    logger.error(e, 'Unexpected exception occurred')
+    logger.error(e,
+      `Unexpected exception occurred. Received payload from Adyen: ${body}`)
     return httpUtils.sendResponse(response, 500)
   }
 }
--- a/notification/src/config/config.js
+++ b/notification/src/config/config.js
@@ -22,7 +22,7 @@ function getFileConfig () {
   let fileConfig = {}
   try {
     fileConfig = require(configPath) // eslint-disable-line
-  } catch (e) {
+  } catch (err) {
     // config file was not provided
   }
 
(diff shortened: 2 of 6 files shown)


notification/src/api/notification/notification.controller.js
```diff
@@ -21,7 +21,8 @@ async function handleNotification (request, response) {
       { 'Content-Type': 'application/json' },
       JSON.stringify({ notificationResponse: '[accepted]' }))
   } catch (e) {
-    logger.error(e, 'Unexpected exception occurred')
+    logger.error(e,
+      `Unexpected exception occurred. Received payload from Adyen: ${body}`)
```
butenkor: It might be better to see body as JSON and not as an escaped JSON text. Did you try:
```
const customErr = { err, "adyenRequestBody": body}
logger.error({customErr}, "Unexpected exception occurred")
```

LEQADA: We lose stacktrace in this case. It looks like we should implement a custom serializer to put it in a separate field. Didn't find an easy solution. Let's keep it as it is?

LEQADA: https://github.com/commercetools/commercetools-adyen-integration/pull/53/commits/99aca3f00c474a8c251cf7ef2c63d1e20ce9b6d0

PR #66, opened 2019-06-11
Change all interfaceId fields to use custom.fields.merchantReference instead

### LINKED ISSUE #62: Extension incompatible with /me/payments endpoint
When trying to use this extension alongside the /me/payments endpoint in order to create and process a payment, the extension throws an error due to being unable to specify an interfaceId from the My Payments version of the endpoint. Did some quick testing and removed the interfaceId validation from the extension and generated a random UUID instead and then the payment processes fine. What would need to be done in order to get this working properly with the my payment endpoint? It seems like you'd want to be able to specify the interfaceId from the client side in order to get the benefit of de-duplicating payments.

zackpollard:
This resolves #62 

This work should change the reference field from interfaceId to a custom field specified on the payment object. If you could give some feedback on whether this is the correct approach, that'd be great. If you want the name of the field to be different too, I can make that change.

All the unit tests past, one test fails but that seems to be failing on master too right now so it seems to be unrelated.

### DIFF
--- a/extension/docs/IntegrationGuide.md
+++ b/extension/docs/IntegrationGuide.md
@@ -102,7 +102,7 @@ or by asynchronous process like [commercetools-payment-to-order-processor job](h
 ### Validate payment
 There must be at least one CTP payment object of type Adyen
 (`Payment.paymentMethodInfo.paymentInterface = ctp-adyen-integration`)
-AND this CTP payment must have `Payment.custom.fields.interfaceId`.
+AND this CTP payment must have `Payment.custom.fields.merchantReference`.
 
 ### Validate payment transaction
 Cart's payment counts as successful if there is at least one payment object
--- a/extension/docs/IntegrationGuide.md
+++ b/extension/docs/IntegrationGuide.md
@@ -103,6 +103,7 @@ or by asynchronous process like [commercetools-payment-to-order-processor job](h
 There must be at least one CTP payment object of type Adyen
 (`Payment.paymentMethodInfo.paymentInterface = ctp-adyen-integration`)
 AND this CTP payment must have `Payment.custom.fields.merchantReference`.
+The combination of `Payment.custom.fields.merchantReference` and `Payment.paymentMethodInfo.paymentInterface` must be unique across the project. This field must not be changed once set.
 
 ### Validate payment transaction
 Cart's payment counts as successful if there is at least one payment object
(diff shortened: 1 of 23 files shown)


extension/resources/payment-custom-types.json
```diff
@@ -7,6 +7,17 @@
     "payment"
   ],
   "fieldDefinitions": [
+    {
+      "type": {
+        "name": "String"
+      },
+      "name": "interfaceId",
```
butenkor: Please rename custom field `interfaceId` to `reference` or `merchantReference` as `interfaceId` has actually different meaning and purpose. It is a unique payment ID which is provided by the payment provider and not the shop itself. Adyen unfortunately does have any generalised payment specific unique IDs but rather transaction specific one only.

extension/test/fixtures/payment-no-method.json
```diff
@@ -15,6 +14,7 @@
       "key": "ctp-adyen-integration-payment-custom-type"
     },
     "fields": {
+      "interfaceId": "pspReference",
```
butenkor: I know it was like that before but `pspReference` implies that it is provided by PSP which is not the case here, right? Maybe better `paymentNoMethodReference` or just the same as in other fixtures `paymentReferenceId`?

notification/test/resources/payment-custom-types.json
```diff
@@ -7,6 +7,17 @@
     "payment"
   ],
   "fieldDefinitions": [
+    {
+      "type": {
+        "name": "String"
+      },
+      "name": "interfaceId",
```
butenkor: Same here.

zackpollard: This has now been changed

extension/resources/payment-custom-types.json
zackpollard: This has now been changed

extension/test/fixtures/payment-no-method.json
zackpollard: Have changed this to paymentReferenceId like in the rest of the tests

extension/docs/IntegrationGuide.md
```diff
@@ -102,7 +102,7 @@ or by asynchronous process like [commercetools-payment-to-order-processor job](h
 ### Validate payment
 There must be at least one CTP payment object of type Adyen
 (`Payment.paymentMethodInfo.paymentInterface = ctp-adyen-integration`)
-AND this CTP payment must have `Payment.interfaceId`.
+AND this CTP payment must have `Payment.custom.fields.merchantReference`.
```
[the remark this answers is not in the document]
zackpollard: Have now added this, thanks :)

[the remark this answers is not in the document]
lojzatran: Thanks for the PR :)

LEQADA:
Thank you for the PR @zackpollard . GJ! 🎉

Issue #65, opened 2019-05-28
Running extension and notification module in AWS Lambda


zackpollard:
Currently both the extension and notification module have to be run as an HTTP endpoint, however the commercetools platform provides the option to run API extensions on AWS Lambda. We would like to use this option if possible, however this would require some changes to the way the modules currently work. Is this something you are looking to eventually build for these modules? If not, would it be something you would be willing to accept a pull request for? If neither of these are an option we will likely end up building a wrapper around the module to call the internal HTTP endpoint, however we would prefer to have a more integrated solution.

zackpollard:
Following from conversations we have had, we are now looking at picking up this piece of work. Is this alright with your team and have you got any input into the way you would prefer this to be built before we go ahead and do the work? We are going to start looking into how we would do this today as it is now a priority for us.

LEQADA:
Thank you for the issue @zackpollard. It's a good idea to make it serverless. 
Could you share the concept of how you see the integration in a serverless environment?

zackpollard:
@LEQADA We are currently halfway through the implementation of this. What we were envisioning is that the lambda will be invoked and things such as ensureResources will run on the first cold start of the lambda, but won't run on following invocations (for that instance of the lambda). We have currently abstracted the logic out of the payment controller and into a payment handler, the payment controller will then just map the HTTP request to what the payment handler expects, and the lambda will have a similar class doing the same thing. We should have a PR ready sometime next week I imagine, I can create a draft PR for now if you would like to keep track of the ongoing work that we are doing. This is just for the extension module right now, but we will do much of the same work over in the notification module once we are done with the extension.

LEQADA:
Sounds good! No need for a draft. Thank you again for initiative. 🙂

zackpollard:
Both of these modules now support running inside a lambda, so i'm going to close this issue 😃

PR #68, opened 2019-06-18
Lambda support for extension module

Max-John:
Lambda support for payment extension module


Max-John:
Hi guys,

Please throw stones at this and suggest all changes you would like. We have tried to limit the changes as much as possible and keep to your conventions.

Max-John:
Thanks for the fast response guys! Just a quick question, but let me know if I should open an issue or if there's a better platform to ask this: Would you consider publishing this via an npm package? We're just trying to work out how to manage our CI/CD and an NPM package seems like a good solution for us. Let us know your thoughts.

butenkor:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/65

butenkor:
> Thanks for the fast response guys! Just a quick question, but let me know if I should open an issue or if there's a better platform to ask this: Would you consider publishing this via an npm package? We're just trying to work out how to manage our CI/CD and an NPM package seems like a good solution for us. Let us know your thoughts.

We will check the option of publishing to npm and let you know.

extension/src/lambda.js
```diff
@@ -0,0 +1,25 @@
...
+const paymentHandler = require('../src/paymentHandler/payment-handler')
+const setup = require('./config/init/ensure-resources')
+
+const logger = utils.getLogger()
+let init = false;
+
+exports.handler = async function(event) {
+  try {
+
+    init = init || await setup.ensureResources();
```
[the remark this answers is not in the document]
zackpollard: We realised this wouldn't work too, this has now been updated to only run once on every cold start

zackpollard:
> Fixes: #65

This doesn't fix #65 on it's own, as this is the extension module only. We are currently working on doing the same for the notification module, should have a PR for that by the end of today.

extension/src/init.js
```diff
@@ -5,9 +5,6 @@ const configLoader = require('./config/config')
 const config = configLoader.load()
 
 const { ensureResources } = require('./config/init/ensure-resources')
-const configLoader = require('./config/config')
-
-const config = configLoader.load()
```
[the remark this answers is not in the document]
zackpollard: Just above that in the file, for some reason it was being loaded twice in the init.js, we saw this and cleaned that up at the same time :)

extension/src/config/init/ensure-resources.js
```diff
@@ -4,16 +4,23 @@ const { ensureApiExtensions } = require('./ensure-api-extensions')
...
     ensurePaymentCustomType(ctpClient),
-    ensureInterfaceInteractionCustomType(ctpClient),
+    ensureInterfaceInteractionCustomType(ctpClient)
+  ])
+}
+
+function ensureResources () {
+  return Promise.all([
+    ensureCustomTypes(),
     ensureApiExtensions(ctpClient, config.load().apiExtensionBaseUrl)
```
butenkor: Looking at this it seems that there might be a case where integration would run into unhandled TypeError [here](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/src/validator/validator-builder.js#L94) if custom was not defined on the payment object. Which might be the case when API extension is created and kicks in before types has been created or when the user of the integration simply forgets to set custom fields on the payment. I guess the issue is not really related to the order of creation here but rather of the linked implementation which misses additional checks?

extension/src/lambda.js
```diff
@@ -0,0 +1,27 @@
...
+const paymentHandler = require('../src/paymentHandler/payment-handler')
+const setup = require('./config/init/ensure-resources')
+
+const logger = utils.getLogger()
+let initialised = false
+
+exports.handler = async function(event) {
+  try {
+    if (!initialised) {
+      await setup.ensureCustomTypes()
```
butenkor: Why would you ensure only types here and not both types and API extension?

zackpollard: In the case of the lambda, we don't want to ensure the API extension as that creates an HTTP API extension link in the commercetools project. For the lambda, it will never be invoked unless that link already exists, therefore making this ensureApiExtension part redundant.

extension/src/lambda.js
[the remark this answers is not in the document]
butenkor: Next please please link to the commit - easier to review :)

extension/src/config/init/ensure-resources.js
zackpollard: Yes, I agree that the hasMerchantReference check should do additional checks to ensure those objects exist. Would you like me to quickly fix that in a separate PR? I don't think this code specifically is affected by that, rather that it was something missed in the PR of the initial implementation of the merchantReference custom field.

extension/src/lambda.js
```diff
@@ -0,0 +1,25 @@
...
+    const paymentResult = await paymentHandler.handlePayment(event.resource.obj)
+
+    return {
+      responseType: paymentResult.success ? `UpdateRequest` : `FailedValidation`,
+      errors: paymentResult.data.errors,
+      actions: paymentResult.data.actions
+    }
+  }
+  catch (e) {
+    logger.error(e, `Unexpected error when processing event ${JSON.stringify(event)}`)
```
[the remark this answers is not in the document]
zackpollard: We were thinking that the benefit would probably be that the exception is coupled to the error message, rather than being two separate entries in cloudwatch that you have to link together manually.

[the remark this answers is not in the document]
Max-John: Yeah good point. The only thing I can think is if we include it there is a single entry in the logs which contains the event and the error together. But you will still have the error duplicated so not great. Happy to remove.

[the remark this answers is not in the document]
butenkor: I actually like that the log holds all the information together. I thought maybe we could avoid throwing after log but I guess it is not an option? I am ok if we keep it like that for now too.

extension/src/config/init/ensure-resources.js
butenkor: Sure, If you have time for that :)

zackpollard: Done, PR here #70 :)

extension/src/lambda.js
butenkor: I see. So in case of Lambda usage it needs to be documented that Commercetoosl API extension needs to be created manually? Or you see any way to automate that?

extension/src/lambda.js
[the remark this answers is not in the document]
Max-John: Yes I guess if it's unexpected exception that we haven't gracefully handled and we agree we want to  fail the actual lambda execution, then I guess we would have to re-throw it. So we'll leave it as is you think?

extension/src/lambda.js
```diff
@@ -0,0 +1,25 @@
...
+
+exports.handler = async function(event) {
+  try {
+
+    init = init || await setup.ensureResources();
+    
+    const paymentResult = await paymentHandler.handlePayment(event.resource.obj)
+
+    return {
+      responseType: paymentResult.success ? `UpdateRequest` : `FailedValidation`,
```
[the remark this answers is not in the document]
Max-John: Hi,
Can you advise in which file we should document this? Or if we should just add a new one for Lambda responses (I don't see anything for the http endpoint so trying to find best place). Thanks

extension/src/lambda.js
zackpollard: So we just had another discussion about this, and we realised that it probably doesn't make sense to create these resources within the lambda init at all. The issue is that these resources all need to exist before the lambda is invoked, so it probably makes sense to not have this ensureCustomTypes step here at all. We weren't actually planning on using this anyway as we will be managing all of our custom objects using terraform with this provider from [here](https://github.com/labd/terraform-provider-commercetools). If we were to remove this step from here, we could just document that if using the lambda, you have to create the resources yourself or using some other kind of automation (such as the terraform provider above). What are your thoughts on this?

extension/test/unit/lambda.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  })
+
+  it('returns correct success response', async () => {
+    const actions = [ { some: "action" }]
+    sinon.stub(paymentHandler, 'handlePayment').returns({ success: true, data: { actions: actions}})
+
+    const result = await handler(event)
+
+    expect(result.responseType).equals(`UpdateRequest`)
+    expect(result.actions).equals(actions)
```
lojzatran: Also check if `result.errors` is empty?

extension/test/unit/lambda.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  })
+
+  it('returns correct failed response', async () => {
+    const errors = [ { some: "error" }]
+    sinon.stub(paymentHandler, 'handlePayment').returns({ success: false, data: { errors: errors}})
+
+    const result = await handler(event)
+
+    expect(result.responseType).equals(`FailedValidation`)
+    expect(result.errors).equals(errors)
```
lojzatran: Also check if `result.actions` is empty?

extension/docs/DevelopmentAndDeploymentGuide.md
```diff
@@ -51,15 +51,21 @@ There are 3 different types of tests. Don't forget to provide all required envir
 with the UI (e.g. credit card 3ds). In such cases, we use Cypress.io. Run `npm run cypress-ui` to test.
 
 ## Deployment
+
+### AWS Lambda
+
+For deployment to lambda zip the extensions folder and specify the src/lambda.handler as the entry point for the function
```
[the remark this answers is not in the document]
zackpollard: This wouldn't be correct, you have to specify the exported handler method as the entrypoint, so src/lambda.handler is correct. I will remove the "the" and encase it in backticks though.

extension/src/lambda.js
```diff
@@ -0,0 +1,27 @@
+const utils = require('./utils')
+const paymentHandler = require('../src/paymentHandler/payment-handler')
+const setup = require('./config/init/ensure-resources')
+
+const logger = utils.getLogger()
+let initialised = false
```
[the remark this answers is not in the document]
zackpollard: That's partly true. When lambdas are run, they do save state for a time, they are just frozen until the next execution. On a "cold start" they have no state, and so will run this code, however on a "warm start" they will still have that state and so this won't be invoked on every run.

However, as I mentioned in another comment on this PR, I don't think it actually makes sense to ensure resources at all within the lambda execution, as if you're doing that at this point, it is already to late. I believe it would be better to force users to create and manage the types themselves if they decide to go with the lambda approach, this could be assisted through automation with terraform.

extension/test/unit/lambda.spec.js
zackpollard: Done

extension/test/unit/lambda.spec.js
zackpollard: Done

extension/docs/DevelopmentAndDeploymentGuide.md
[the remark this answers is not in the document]
zackpollard: Done

zackpollard:
> Can you run `npm run lint` and check for the errors there?

I have run the linting and committed all the changes.

extension/src/lambda.js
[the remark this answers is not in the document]
butenkor: ok

extension/src/lambda.js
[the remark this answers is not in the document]
butenkor: @Max-John I guess it is not necessary as responseType is being seen and evaluated by Commercetools API only so the caller will rather see an error returned by API, which structure is described [here](https://docs.commercetools.com/http-api-projects-api-extensions#the-api-extension-within-the-flow-of-the-api-call). Maybe we could link as separate point under https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/IntegrationGuide.md#error-cases:

`For more info on API-Extensions error handling please see [here](https://docs.commercetools.com/http-api-projects-api-extensions#the-api-extension-within-the-flow-of-the-api-call)`

Additionally if you have any input it could be useful to extend our best practices into 2 sections ("AWS Lambda" and "Hosted deployment"):
https://github.com/commercetools/commercetools-adyen-integration/blob/master/docs/BEST_PRACTICES.md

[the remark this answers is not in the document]
butenkor: @lojzatran Do you know why we did not provide in response `responseType` before? According to API doc it seem mandatory: https://docs.commercetools.com/http-api-projects-api-extensions#the-api-extension-within-the-flow-of-the-api-call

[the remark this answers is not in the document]
zackpollard: @butenkor With HTTP extensions the responseType is taken care of using HTTP status codes 200, 201 and 400. Lambdas only need this as they have no concept of status code responses.

extension/src/lambda.js
zackpollard: @butenkor Do you have any thoughts on the above?

butenkor: > So we just had another discussion about this, and we realised that it probably doesn't make sense to create these resources within the lambda init at all. The issue is that these resources all need to exist before the lambda is invoked, so it probably makes sense to not have this ensureCustomTypes step here at all.

Agree.

> If we were to remove this step from here, we could just document that if using the lambda, you have to create the resources yourself or using some other kind of automation (such as the terraform provider above). What are your thoughts on this?

Maybe for convenience we could add terraform template (for custom types and API extension) here and a short guide how to run it?

Alternatively it could be just a simple function which can be run from command line and create required resources (due to complexity updates on the type definition would not be supported for now).

extension/src/lambda.js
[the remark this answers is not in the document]
butenkor: I see. Thanks

extension/src/lambda.js
zackpollard: I have now removed the ensureCustomTypes from the lambda.

I have create a seperate npm command that can be used to invoke the ensureCustomTypes method from ensure-resources.js. If provided with the CTP project key, client id and secret, it will create the custom types in the project. I have also added this detail to the documentation under the lambda section.

extension/docs/DevelopmentAndDeploymentGuide.md
```diff
@@ -56,6 +56,9 @@ with the UI (e.g. credit card 3ds). In such cases, we use Cypress.io. Run `npm r
 
 For deployment to lambda zip the extensions folder and specify `src/lambda.handler` as the entry point for the function
 
+When deploying the lambda, it will not create the custom types for you. These are required for the extension to operate correctly.
+You can create these by running the command `npm run create-custom-types` and providing the `CTP_PROJECT_KEY`, `CTP_CLIENT_ID` and `CTP_CLIENT_SECRET` environment variables.
+
```
butenkor: Please also add an info that API extension needs to be created in CT manually.

extension/src/lambda.js
butenkor: Besides the 1 comment i left in one of the latest your commits, looks good. Thanks!

extension/docs/DevelopmentAndDeploymentGuide.md
zackpollard: Yep, absolutely, done!

zackpollard:
@butenkor @lojzatran So I have come across a problem now that we have removed the ensureResources calls for the lambda. This now means that the lambda doesn't require environment variables for CTP project credentials. However, in config.js, on load it asserts that all the CTP project credentials exist, and if they don't then it throws an error. Do you have any suggestions as to what to do with this now, as there's no need to assert that in the case of the lambda. I'm thinking I could move that check to the init.js so that it is only checked on the server load. What do you think?

extension/src/lambda.js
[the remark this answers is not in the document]
lojzatran: So you will remove this part with `initialised`?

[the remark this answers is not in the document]
lojzatran: Ah I see now, it's already removed.

zackpollard:
> @butenkor @lojzatran So I have come across a problem now that we have removed the ensureResources calls for the lambda. This now means that the lambda doesn't require environment variables for CTP project credentials. However, in config.js, on load it asserts that all the CTP project credentials exist, and if they don't then it throws an error. Do you have any suggestions as to what to do with this now, as there's no need to assert that in the case of the lambda. I'm thinking I could move that check to the init.js so that it is only checked on the server load. What do you think?

I have now done what I suggested here, please see the latest commit for the simple code change

zackpollard:
I have updated the documentation contents using doctoc for this PR too as it was mentioned on the notification module PR

extension/docs/DevelopmentAndDeploymentGuide.md
```diff
@@ -51,15 +53,28 @@ There are 3 different types of tests. Don't forget to provide all required envir
 with the UI (e.g. credit card 3ds). In such cases, we use Cypress.io. Run `npm run cypress-ui` to test.
 
 ## Deployment
+
+### AWS Lambda
+
+For deployment to lambda zip the extensions folder and specify `src/lambda.handler` as the entry point for the function
+
+When deploying the lambda, it will not create the custom types for you. These are required for the extension to operate correctly.
```
lojzatran:
```suggestion
When deploying the lambda, it will NOT create the custom types for you. These are required for the extension to operate correctly. Please add [payment custom types](../resources/payment-custom-types.json) and [payment interface interaction types](../resources/payment-interface-interaction-types.json) manually.
```

lojzatran: @zackpollard can you make this small change and check if the links I provided are correct? Then I will approve this PR and merge.

zackpollard: Done!

PR #70, opened 2019-06-19
Fix potential error with merchantReference validation

zackpollard:
Do more checks to ensure the underlying objects exist before checking that the merchantReference field isn't empty

### DIFF
--- a/extension/src/validator/validator-builder.js
+++ b/extension/src/validator/validator-builder.js
@@ -91,7 +91,9 @@ function withPayment (paymentObject) {
       return this
     },
     validateMerchantReferenceField () {
-      const hasMerchantReference = !_.isEmpty(paymentObject.custom.fields.merchantReference)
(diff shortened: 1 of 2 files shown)


LEQADA:
Thank you for the PR @zackpollard . A unit test to reproduce the failing case would be perfect

zackpollard:
> Thank you for the PR @zackpollard . A unit test to reproduce the failing case would be perfect

Agreed, this logic is applied to a lot of different validations in that file however, should I create a unit test for each one to ensure this doesn't happen for any of those in the future too?

LEQADA:
One case should be enough. The logic is the same.

zackpollard:
I've now added a unit test that invokes all those validations with an empty payment object which would have failed before in this case. It's using the same it case but is looping over all of the different validations that could have this potential error case introduced through a degradation

zackpollard:
> Can you run `npm run lint` and check if there are any errors?

Done :)

PR #73, opened 2019-06-24
Ability to disable ensure resources

### LINKED ISSUE #72: Ability to disable resource creation in extension and notification module
It would be good to be able to disable the automatic resource creation/updating that happens as part of the startup of the server for both the extension module and the notification module. I propose that an environment variable boolean flag is provided so that if users are managing these resources themselves (i.e. through terraform) then they have the ability to disable this functionality.

zackpollard:
This fixes #72. 

This is based off of the changes in #68. 

Once that PR is merged, this one should follow. This is due to conflicts that would otherwise exist due to changes made in that PR

### DIFF
--- a/extension/docs/DevelopmentAndDeploymentGuide.md
+++ b/extension/docs/DevelopmentAndDeploymentGuide.md
@@ -36,6 +36,7 @@ Environment variables to configure the notification module:
 |`API_EXTENSION_BASE_URL` | URL of the Extension module server. In case of any payment changes, [CTP API extension](https://docs.commercetools.com/http-api-projects-api-extensions) will call this URL and pass the payment object in body. | YES | |
 |`LOG_LEVEL` | bunyan log level (`trace`, `debug`, `info`, `warn`, `error`, `fatal`)| NO | `info`
 |`PORT` | port on which the application will run | NO | 8080
+|`DISABLE_ENSURE_RESOURCES` | disables the creation of resources in commercetools (e.g. custom types) | NO | `false`
 
 
 After setting all environmental variables, execute command `npm run extension` to run the module.
--- a/extension/docs/DevelopmentAndDeploymentGuide.md
+++ b/extension/docs/DevelopmentAndDeploymentGuide.md
@@ -36,7 +36,7 @@ Environment variables to configure the notification module:
 |`API_EXTENSION_BASE_URL` | URL of the Extension module server. In case of any payment changes, [CTP API extension](https://docs.commercetools.com/http-api-projects-api-extensions) will call this URL and pass the payment object in body. | YES | |
 |`LOG_LEVEL` | bunyan log level (`trace`, `debug`, `info`, `warn`, `error`, `fatal`)| NO | `info`
 |`PORT` | port on which the application will run | NO | 8080
-|`DISABLE_ENSURE_RESOURCES` | disables the creation of resources in commercetools (e.g. custom types) | NO | `false`
+|`ENSURE_RESOURCES` | Set to `false` to disable the creation of resources in commercetools (e.g. custom types) | NO | `true`
 
 
 After setting all environmental variables, execute command `npm run extension` to run the module.
--- a/extension/src/config/config.js
+++ b/extension/src/config/config.js
@@ -7,7 +7,7 @@ function getEnvConfig () {
     apiExtensionBaseUrl: process.env.API_EXTENSION_BASE_URL,
     keepAliveTimeout: !Number.isNaN(process.env.KEEP_ALIVE_TIMEOUT) ? parseFloat(process.env.KEEP_ALIVE_TIMEOUT, 10)
       : undefined,
-    disableEnsureResources: process.env.DISABLE_ENSURE_RESOURCES === 'true' || false
+    ensureResources: process.env.ENSURE_RESOURCES !== 'false'
   }
 }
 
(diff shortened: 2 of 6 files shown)


zackpollard:
Thanks for merging the other two PRs! Now that they are merged, I have reset this one based on master so it only has the two new commits in there. As always, shout if there's anything you want any changes done!

extension/docs/DevelopmentAndDeploymentGuide.md
```diff
@@ -36,6 +36,7 @@ Environment variables to configure the notification module:
 |`API_EXTENSION_BASE_URL` | URL of the Extension module server. In case of any payment changes, [CTP API extension](https://docs.commercetools.com/http-api-projects-api-extensions) will call this URL and pass the payment object in body. | YES | |
 |`LOG_LEVEL` | bunyan log level (`trace`, `debug`, `info`, `warn`, `error`, `fatal`)| NO | `info`
 |`PORT` | port on which the application will run | NO | 8080
+|`DISABLE_ENSURE_RESOURCES` | disables the creation of resources in commercetools (e.g. custom types) | NO | `false`
```
butenkor: Negations in variable names are confusing to read once you need to negate those in `if` statements. How about:
`ENSURE_RESOURCES` with default `true`?

extension/src/config/config.js
```diff
@@ -6,7 +6,8 @@ function getEnvConfig () {
     logLevel: process.env.LOG_LEVEL,
     apiExtensionBaseUrl: process.env.API_EXTENSION_BASE_URL,
     keepAliveTimeout: !Number.isNaN(process.env.KEEP_ALIVE_TIMEOUT) ? parseFloat(process.env.KEEP_ALIVE_TIMEOUT, 10)
-      : undefined
+      : undefined,
+    disableEnsureResources: process.env.DISABLE_ENSURE_RESOURCES === 'true' || false
```
butenkor: Why not use here same naming as the env name itself?

extension/src/config/init/ensure-resources.js
```diff
@@ -2,8 +2,9 @@ const { ensurePaymentCustomType } = require('./ensure-payment-custom-type')
 const { ensureInterfaceInteractionCustomType } = require('./ensure-interface-interaction-custom-type')
 const { ensureApiExtensions } = require('./ensure-api-extensions')
 const ctpClientBuilder = require('../../ctp/ctp-client')
-const config = require('../../config/config')
+const configLoader = require('../../config/config')
 
+const config = configLoader.load()
 const ctpClient = ctpClientBuilder.get()
 
 function ensureCustomTypes () {
```
butenkor: This should not be disabled too as it is exposed for public usage?

extension/src/config/init/ensure-resources.js
```diff
@@ -14,9 +15,10 @@ function ensureCustomTypes () {
 }
 
 function ensureResources () {
+  if (config.disableEnsureResources) return Promise.resolve()
```
butenkor: Why not have this statement where ensure resources is being called? Seems confusing if you see in one place that function `ensureResources` is called but internally does not do that, wdyt?

extension/docs/DevelopmentAndDeploymentGuide.md
zackpollard: I think it makes it more explicit that you are specifically disabling the ensure resources portion of the application.

extension/src/config/config.js
zackpollard: I'm not sure what you mean, as far as I can see it is the same, just camelcase.

extension/src/config/init/ensure-resources.js
zackpollard: This shouldn't be disabled because it is either called from the ensureResources method, or is being specifically invoked via the npm command that runs this method in which case, would you really want to disable it if they passed this variable in? Why would they call this method (using the npm command) to create resources and also want to disable resource creation?

extension/src/config/init/ensure-resources.js
zackpollard: I prefer it here personally, as I don't think the caller of this method should worry about whether the resources actually get created or not. The end result should be the same either way, the resources exist. By disabling ensure resources, you are saying that you are managing them yourself essentially, and so they should still exist.

extension/docs/DevelopmentAndDeploymentGuide.md
butenkor: I get the point but it is just not a good practice in general. Just some examples:
http://truelogic.org/wordpress/2009/02/27/how-not-to-name-a-boolean-variable/
http://davej.io/2018/05/negative-naming.html
https://softwareengineering.stackexchange.com/questions/196830/boolean-method-naming-affirmative-vs-negative

extension/src/config/init/ensure-resources.js
butenkor: I see. Thanks

extension/docs/DevelopmentAndDeploymentGuide.md
zackpollard: Have changed this to `ENSURE_RESOURCES` now with default of `true`

butenkor: Thanks!

PR #76, opened 2019-07-24
Restrict client scope

zackpollard:
This PR will restrict the scope that the notification module uses to request client credentials to only what is required instead of always using manage_project for everything. Security best practice means we should not be providing all access tokens to services that don't need them.

This code will always grant `manage_payments`, will grant `manage_types` if `ENSURE_RESOURCES` isn't set to `false` and will grant `manage_extensions` when running integration tests so that they can perform their destroy and create of the ngrok based extension before running each test.

This code will be a breaking change for any deployments that currently use `manage_project` scope as commercetools will not let you request a lesser scope such as `manage_payments` if your token was created with `manage_project`.

As always, please let me know if there's anything that needs changing/adding!

### DIFF
--- a/notification/test/integration/notification.handler.spec.js
+++ b/notification/test/integration/notification.handler.spec.js
@@ -12,7 +12,7 @@ const notifications = require('../resources/notification')
 const localhostIp = address()
 
 describe('notification module', () => {
-  const ctpClient = ctpClientBuilder.get(config)
+  const ctpClient = ctpClientBuilder.get(config, ['manage_extensions:'.concat(config.ctp.projectKey)])
 
   before(async () => {
     await iTSetUp.startServer()
--- a/notification/src/utils/ctp.js
+++ b/notification/src/utils/ctp.js
@@ -17,7 +17,7 @@ function createCtpClient (config, extraScopes) {
 
   let scopeBuilder = [scopes.MANAGE_PAYMENTS.concat(':').concat(projectKey)]
   if (config.ensureResources) scopeBuilder = scopeBuilder.concat(scopes.MANAGE_TYPES.concat(':').concat(projectKey))
-  if (extraScopes) scopeBuilder = scopeBuilder.concat('manage_extensions:'.concat(projectKey))
+  if (extraScopes) scopeBuilder = scopeBuilder.concat(extraScopes)
 
   const authMiddleware = createAuthMiddlewareForClientCredentialsFlow({
     host: AUTH_HOST,
--- a/notification/src/utils/ctp.js
+++ b/notification/src/utils/ctp.js
@@ -5,12 +5,20 @@ const { createAuthMiddlewareForClientCredentialsFlow } = require('@commercetools
 const { createHttpMiddleware } = require('@commercetools/sdk-middleware-http')
 const { createQueueMiddleware } = require('@commercetools/sdk-middleware-queue')
 const { createRequestBuilder } = require('@commercetools/api-request-builder')
+const { scopes } = require('@commercetools/sdk-middleware-auth')
 
-function createCtpClient ({
-  clientId, clientSecret, projectKey, concurrency = 10
-}) {
+function createCtpClient (config, extraScopes) {
   const AUTH_HOST = 'https://auth.commercetools.com'
   const API_HOST = 'https://api.commercetools.com'
+  const clientId = config.ctp.clientId
+  const clientSecret = config.ctp.clientSecret
+  const projectKey = config.ctp.projectKey
+  const concurrency = 10
+
+  let scopeBuilder = [scopes.MANAGE_PAYMENTS.concat(':').concat(projectKey)]
+  if (config.ensureResources) scopeBuilder = scopeBuilder.concat(scopes.MANAGE_TYPES.concat(':').concat(projectKey))
+  if (extraScopes) scopeBuilder = scopeBuilder.concat('manage_extensions:'.concat(projectKey))
+
   const authMiddleware = createAuthMiddlewareForClientCredentialsFlow({
     host: AUTH_HOST,
     projectKey,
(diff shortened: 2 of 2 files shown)


zackpollard:
Pushed a small fix to this so it actually uses the list of extra scopes that is passed in rather than a hardcoded scope... woops.

Is it possible to get some reviews done on this as we would like to get this merged and start using it soon if possible. Cheers!

butenkor:
> Security best practice means we should not be providing all access tokens to services that don't need them

That is in general correct but i am not sure how does it help if the app itself might have [manage_project access](https://github.com/commercetools/commercetools-adyen-integration/pull/76/files#diff-84971f8d0caa90691446ce9e952d9366R14) anyway? Maybe you should rather create a dedicated CTP API client in merchant center which has limited scope instead?

Also keep i mind that required scope here will likely change soon as due to Adyen's risk management [checks](https://docs.adyen.com/risk-management/provide-data-for-risk-checks) as also various other [payment modes](https://docs.adyen.com/classic-integration/hosted-payment-pages/klarna-afterpay-and-ratepay-open-invoice#ratepay-installments) we will need access to both cart and the customer too.

zackpollard:
> Maybe instead you should rather create a dedicated CTP API client in merchant center with limited scope?

This is what we want to do, but currently the notification module requests manage_project and therefore creating a client with restricted scope causes the module to fail.



> Also keep in mind that required client scope will likely change soon as due to Adyen's risk management [checks](https://docs.adyen.com/risk-management/provide-data-for-risk-checks) as also various other [payment modes](https://docs.adyen.com/classic-integration/hosted-payment-pages/klarna-afterpay-and-ratepay-open-invoice#ratepay-installments) we will need access to both the customer and the cart.

Personally I think that's fine, and those scopes should be added as and when they are required, and new client credentials should be created.


It doesn't make sense to give this notification service permissions to access all data on the system with manage_project, it's a huge security risk.

notification/test/integration/notification.handler.spec.js
```diff
@@ -12,7 +12,7 @@ const notifications = require('../resources/notification')
 const localhostIp = address()
 
 describe('notification module', () => {
-  const ctpClient = ctpClientBuilder.get(config)
+  const ctpClient = ctpClientBuilder.get(config, ['manage_extensions:'.concat(config.ctp.projectKey)])
```
butenkor: Notification module does not need to manage extensions. This is required for extension module only.

zackpollard: It does because of line 26 on integration-test-set-up.js which does a delete on all payments, types and extensions to ensure there is no interference with the integration tests.

butenkor:
> This is what we want to do, but currently the notification module requests manage_project and therefore creating a client with restricted scope causes the module to fail.

Where does it require manage_project? It takes whatever scope is supplied by provided client credentials. If provided client credentials scope is manage_project it will also generate by default an access token with the same scope. Currently it would be enough if you would supply a client over env vars which covers following scopes: manage_types, manage_payments, manage_extensions. This will ensure that only mentioned scopes are available for the whole application and no change to the code is required.

> It doesn't make sense to give this notification service permissions to access all data on the system with manage_project, it's a huge security risk.

Agree. Thats why you can supply credentials with limited scope already. It just needs to be documented which scopes are required for module to properly work.

I wonder how current change improves security? In case of client passed with manage_project scope we would just programatically request less but app itself could potentially still use higher scope. Also according to the doc you can not request less scope then provided by the client credentials: https://docs.commercetools.com/http-api-scopes#scopes

zackpollard:
> Where does it require manage_project?

It doesn't require any scopes, that is the problem. In the commercetools-sdk-middleware-auth library, if you don't specify any scopes, it defaults to using manage_project. There is no way for it to lookup what scopes the client_secret and client_id have programatically, so it just assumes manage_project. These changes ensure that doesn't happen, and specifies what scopes the library should use when creating the client credentials. You can see the code that does that (here)[https://github.com/commercetools/nodejs/blob/1731a8e325135590a64e0540ad9f17cba64c90d9/packages/sdk-middleware-auth/src/build-requests.js#L36].

butenkor:
Then it seems that an issue resides within the SDK as it should not default to manage_project. We will create an issue for it. According to the doc, scope is optional and API will automatically apply the scopes of the provided client so it should not be necessary for you to guess them. On the other hand your fix will not work for supplied credentials with manage_project scope as you can not request less scope than client provides.

lojzatran:
Hi @zackpollard, thanks for this PR. We created an issue for nodejs team: https://github.com/commercetools/nodejs/issues/1382

zackpollard:
> Then it seems that an issue resides within the SDK as it should not default to manage_project. We will create an issue for it. According to the doc, scope is optional and API will automatically apply the scopes of the provided client so it should not be necessary for you to guess them. On the other hand your fix will not work for supplied credentials with manage_project scope as you can not request less scope than client provides.

Alright, I did not know that providing no scope was an option. If that is possible, then that definitely seems like the right approach. Also, I noted that with these changes using a manage_project scoped token wouldn't be possible, which would make it a breaking change with anyone who currently has it deployed. But also deploying it with manage_project is bad so they should probably stop doing that anyway 😉 

Hopefully the nodejs team get this fixed soon and then those changes can be pulled in here with a new version release :)

Cheers for your help!

zackpollard:
@lojzatran @butenkor This has now been fixed in upstream. Is it possible to get a new release out using that new version of the library to fix this problem?

Thanks!

butenkor:
@zackpollard it is released. Let us know if this PR can be closed.
https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v2.1.1

zackpollard:
Yea I see no reason to leave this open now. I haven't had a chance to test it with this method yet, but will get around to that shortly. If there are any issues I will open an issue instead. Cheers!

Issue #231, opened 2020-03-31
Add support for getPaymentMethods and getOriginKeys


ahmetoz:
Verify before processing the methodInfoInterface that payment. 
Add support for getting getPaymentMethods and getOriginKeys,  log data within interface interaction and provide response to the user within getPaymentMethodsResponse and getOriginKeyResponse custom fields.

Issue #75, opened 2019-07-23
createdAt on addInterfaceInteraction action is empty


lojzatran:
We got this error in a notification module where `createdAt={}`. This returns error from the platform:
```json
{
  "code": 502,
  "statusCode": 502,
  "status": 502,
  "message": "'500' is not an expected status code.",
  "originalRequest": {
    "uri": "/our-demo-project/payments/9ff06372-4812-4293-916b-01c58864c056",
    "method": "POST",
    "body": {
      "version": 101,
      "actions": [
        {
          "action": "addInterfaceInteraction",
          "type": {
            "key": "ctp-adyen-integration-interaction-notification",
            "typeId": "type"
          },
          "fields": {
            "createdAt": {},
            "status": "AUTHORISATION",
            "notification": "{\"NotificationRequestItem\":{\"additionalData\":{\"expiryDate\":\"10/2020\",\"authCode\":\"39697\",\"cardSummary\":\"1142\"},\"amount\":{\"currency\":\"EUR\",\"value\":495},\"eventCode\":\"AUTHORISATION\",\"eventDate\":\"2019-07-03T19:19:30+02:00\",\"merchantAccountCode\":\"CommercetoolsGmbHDE775\",\"merchantReference\":\"paymentReferenceId\",\"operations\":[\"CANCEL\",\"CAPTURE\",\"REFUND\"],\"paymentMethod\":\"visa\",\"pspReference\":\"853562174370815K\",\"reason\":\"39697:1142:10/2020\",\"success\":\"true\"}}"
          }
        }
      ]
    },
    "headers": {
      "Accept": "application/json",
      "Content-Type": "application/json",
      "Authorization": [
        "Bearer ********"
      ]
    }
  },
  "retryCount": 0,
  "headers": {
    "server": [
      "nginx"
    ],
    "date": [
      "Fri, 05 Jul 2019 22:04:19 GMT"
    ],
    "content-type": [
      "application/json; charset=utf-8"
    ],
    "content-length": [
      "337"
    ],
    "x-http-status-caused-by-extension": [
      "true"
    ],
    "server-timing": [
      "projects;dur=74"
    ],
    "x-correlation-id": [
      "projects-c055bbda-12ee-45d2-9b53-09e9f02fc4a4"
    ],
    "x-served-by": [
      "api-gce-bf772e57.sphere.prod.commercetools.de"
    ],
    "x-served-config": [
      "sphere-projects-ws-1.0"
    ],
[...]

Issue #46, opened 2019-04-04
In case of Node.js errors save an interfaceInteraction so the shop knows what happened to a payment


LEQADA:
Right now we save interfaceInteraction only if we get a response from the request to Adyen. If any Node.js error happens the error will be catchen in the controller and logged. 
This way it's hard for a shop to understand what happened to his payment. He needs to dive into logs.

PR #232, opened 2020-03-31
Add support for getPaymentMethods and getOriginKeys #231

ahmetoz:
Issue: #231


extension/test/integration/get-origin-keys.handler.spec.js
```diff
@@ -0,0 +1,119 @@
...
+        }
+      }
+    }
+
+    const { statusCode, body: payment } = await ctpClient.create(ctpClient.builder.payments, paymentDraft)
+    expect(statusCode).to.equal(201)
+
+    const { getOriginKeysRequest, getOriginKeysResponse } = payment.custom.fields
+    expect(getOriginKeysRequest).to.be.deep.equal(JSON.stringify(getOriginKeysRequestDraft))
+    expect(JSON.parse(getOriginKeysResponse)).to.be.deep.equal({
```
lojzatran, line 99: I don't think we care about the response in case of error. We care that the key is not set and sth is stored in here. This error can be changed any time and then our test will fail.

extension/test/integration/get-origin-keys.handler.spec.js
```diff
@@ -0,0 +1,119 @@
...
+          key: c.CTP_WEB_COMPONENTS_PAYMENT_TYPE_KEY
+        },
+        fields: {
+          getOriginKeysRequest: JSON.stringify(getOriginKeysRequestDraft)
+        }
+      }
+    }
+
+    const { statusCode, body: payment } = await ctpClient.create(ctpClient.builder.payments, paymentDraft)
+    expect(statusCode).to.equal(201)
```
lojzatran, line 95: What is the reason that we create in the case of error and not throw an error to the platform? Can you add a comment to the task with explanation? We will later add the reason to our docs. https://jira.commercetools.com/browse/CTPI-209

extension/test/integration/get-payment-methods.handler.spec.js
```diff
@@ -0,0 +1,211 @@
...
+      ...getPaymentMethodsRequestDraft
+    })
+
+    const interfaceInteractionResponse = JSON.parse(interfaceInteraction.fields.response)
+    expect(interfaceInteractionResponse.groups).to.be.an.instanceof(Array)
+    expect(interfaceInteractionResponse.paymentMethods).to.be.an.instanceof(Array)
+    expect(interfaceInteractionResponse.additionalData).to.not.exist
+  })
+
+  it('given an existing payment' +
```
lojzatran, line 73: I don't think we need these error cases as integration tests. It's enough to have them as unit test where you mock call to adyen. If we keep adding integration tests, it will take very long time to run in the end.

extension/test/integration/get-origin-keys.handler.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    expect(JSON.parse(interfaceInteraction.fields.request)).to.be.deep.equal({
+      merchantAccount: process.env.ADYEN_MERCHANT_ACCOUNT,
+      ...getOriginKeysRequestDraft
+    })
+
+    const interfaceInteractionResponse = JSON.parse(interfaceInteraction.fields.response)
+    expect(interfaceInteractionResponse.originKeys).to.exist
+  })
+
+  it('given a payment ' +
```
lojzatran, line 69: I don't think we need these error cases as integration tests. It's enough to have them as unit test where you mock call to adyen. If we keep adding integration tests, it will take very long time to run in the end.

extension/src/config/config.js
```diff
@@ -16,17 +16,17 @@ function getCTPEnvCredentials () {
     projectKey: process.env.CTP_PROJECT_KEY,
     clientId: process.env.CTP_CLIENT_ID,
     clientSecret: process.env.CTP_CLIENT_SECRET,
-    apiUrl: 'https://api.sphere.io',
-    authUrl: 'https://auth.sphere.io'
+    apiUrl: process.env.CTP_HOST || 'https://api.sphere.io',
```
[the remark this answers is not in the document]
ahmetoz: do we know the region? I'm not which one is default?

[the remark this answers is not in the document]
lojzatran: I guess it depends where you are. Since we're in Europe, I would choose Frankfurt. Any of them is better than `api.sphere.io`, because it is deprecated.

extension/src/config/config.js
```diff
@@ -16,17 +16,17 @@ function getCTPEnvCredentials () {
...
   }
 }
 
 function getAdyenCredentials () {
   return {
     merchantAccount: process.env.ADYEN_MERCHANT_ACCOUNT,
     apiKey: process.env.ADYEN_API_KEY,
-    apiBaseUrl: process.env.ADYEN_API_BASE_URL || 'https://checkout-test.adyen.com/v40',
-    legacyApiBaseUrl: process.env.ADYEN_LEGACY_API_BASE_URL || 'https://pal-test.adyen.com/pal/servlet/Payment/v40'
+    apiBaseUrl: process.env.ADYEN_API_BASE_URL || 'https://checkout-test.adyen.com/v53',
```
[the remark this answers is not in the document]
ahmetoz: I was testing with v53 on the postman because of their open API spec, actually, I did not check it from the docs. Maybe we could ask Adyen for the difference ?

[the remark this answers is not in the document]
lojzatran: Which spec are you talking about? Can you give a link?

extension/src/paymentHandler/payment-handler.js
```diff
@@ -1,56 +1,42 @@
...
-    return paymentHandlers.kcpHandler
-  return paymentHandlers.fetchPaymentMethodsHandler
+function _getPaymentHandlers (paymentObject) {
+  const result = []
+  if (paymentObject.custom.fields.getOriginKeysRequest && !paymentObject.custom.fields.getOriginKeysResponse)
+    result.push(paymentHandlers.getOriginKeysHandler)
+  if (paymentObject.custom.fields.getPaymentMethodsRequest && !paymentObject.custom.fields.getPaymentMethodsResponse)
+    result.push(paymentHandlers.getPaymentMethodsHandler)
+  return result
 }
```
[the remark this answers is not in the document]
ahmetoz: sorry which line, line 41 ?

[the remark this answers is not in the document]
lojzatran: Yes

extension/src/paymentHandler/payment-utils.js
```diff
@@ -104,11 +104,11 @@ function createChangeTransactionStateAction (transactionId, transactionState) {
   }
 }
 
-function createSetCustomFieldAction (name, value) {
+function createSetCustomFieldAction (name, response) {
   return {
     action: 'setCustomField',
     name,
-    value
+    value: JSON.stringify(response)
```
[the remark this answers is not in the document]
ahmetoz: I see what you mean. I will check it, also I guess we don't have a reason to persist a LocalisedEnum to the custom field.

extension/test/integration/get-origin-keys.handler.spec.js
ahmetoz: > we create in the case of error and not throw an error to the platform?

I am also not sure from that, What will you throw from the platform? 
Because here in that example, the error is an interaction with Adyen. I guess this should not be thrown,
but when you check the other parts we don't have a proper error handling mechanism.
Which I already add those things as a todo, so I could get some feedback because there are a couple of ways to implement those and could be changed with other parts.

https://github.com/commercetools/commercetools-adyen-integration/blob/00a9cba17a27e41703c928d0eef7e7d35f54e6c0/extension/src/paymentHandler/payment-handler.js#L11-L16

Another point in here, I am also not throwing an error it could be the requestString could be invalid, which could be wrong. So, in summary, we could have lots of edge cases to handle errors and we need to discuss those, which that PR could be the best candidate for this.

https://github.com/commercetools/commercetools-adyen-integration/blob/2fad63252f9fb3da0f6faa80dc9b17fd94b8a13d/extension/src/paymentHandler/web-component-service.js#L15-L23

extension/test/integration/get-payment-methods.handler.spec.js
ahmetoz: This project is not handling too many cases on integration tests, so performance is not a problem here yet, and IMO it will not be a problem in the future. also, those tests are not slow. But I could consider adding those as a unit test as you said those parts really depend on the Adyen and unit test could be a better candidate, yes.

extension/src/paymentHandler/payment-utils.js
```diff
@@ -59,25 +59,25 @@ function getMatchingCtpState (adyenState) {
...
   // strip away sensitive data
   delete response.additionalData
 
   return {
     action: 'addInterfaceInteraction',
-    type: { key: c.CTP_INTERFACE_INTERACTION },
+    type: { key: c.CTP_INTERFACE_INTERACTION_PAYMENT_TYPE_KEY },
     fields: {
-      createdAt: new Date(),
+      createdAt: new Date().toISOString(),
```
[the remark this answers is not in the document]
ahmetoz: > I was also doing it, but sometimes I get a weird output, so I changed it

Thanks for that suggestion, I am trying to understand that argument, which looks like your experience with toISOString was not good, I guess your problem could be related to Date() object.. Any idea how we could solve it or any alternative also could be a solution.

##### Why I added it ?

A typical createdAt if I use toISOString() will look sth like: "createdAt": "2020-04-07T07:44:22.132Z"
the values for the payment resource will look like: 
"createdAt": "2020-04-07T07:44:22.249Z",
"lastModifiedAt": "2020-04-07T07:44:22.249Z",

without using iso string, it will look like: "createdAt": "2020-04-06T09:17:19.829Z",
as you would know it will use my local date instead of UTC.

extension/src/paymentHandler/payment-handler.js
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/caaee4d969437a36590402ffb50bf9de5cc5bcbf

extension/src/paymentHandler/payment-handler.js
```diff
@@ -1,56 +1,42 @@
...
-    return paymentHandlers.refundHandler
-  if (paymentValidator.isPaypal())
-    return paymentHandlers.paypalHandler
-  if (paymentValidator.isCreditCard())
-    return paymentHandlers.creditCardHandler
-  if (paymentValidator.isKcp())
-    return paymentHandlers.kcpHandler
-  return paymentHandlers.fetchPaymentMethodsHandler
+function _getPaymentHandlers (paymentObject) {
+  const result = []
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/00a9cba17a27e41703c928d0eef7e7d35f54e6c0

extension/src/config/config.js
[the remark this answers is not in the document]
ahmetoz: for now, the correct number seems to be v52. as all their documents and their code examples are using v52. will change this.

extension/src/config/config.js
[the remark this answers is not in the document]
ahmetoz: Actually it could be Belgium too, so wdyt without using a default value and make it required ?

extension/src/config/config.js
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/e10699da3b521e2857864d1e0f992a1b3db9966a

extension/src/config/config.js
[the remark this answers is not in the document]
lojzatran: I would put sth there by default. If it's too slow, than the user can choose another API.

extension/src/paymentHandler/payment-utils.js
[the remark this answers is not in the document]
lojzatran: If you tested it and it looks ok, then no problem.

extension/src/config/config.js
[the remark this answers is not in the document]
ahmetoz: What is the benefit of having a default value for sth not clear? This is not a batch size so you could not predict it, I would go with the required one so library user should define it like a projectkey and clientId.

extension/test/integration/get-origin-keys.handler.spec.js
lojzatran: It's ok for me too if we discuss about common error handling in a different task. Create a JIRA so that we don't forget if it's not already created.

ahmetoz: No, I meant that:
> which that PR could be the best candidate for this.

lojzatran: Which PR is that PR?

ahmetoz: current PR.

extension/src/config/config.js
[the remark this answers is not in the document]
lojzatran: I'm more for concept of having most of the things by default and the user has to set up minimum stuff.
In this case if you choose Europe and you're in America, it will work anyway without any problems and you don't have to set up. It's not sth that will not work at all. It's just maybe not optimal for some cases. In such situations you change the default.

extension/test/integration/get-payment-methods.handler.spec.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/a48be471fd6b07f2f29f90297e6cc16a7038a94b

I am skipping the last test, as I could not decide yet where to put json invalid validations, for the things which I did not fix in this PR, added as a todo for not to forget later.

extension/test/integration/get-origin-keys.handler.spec.js
ahmetoz: This is the same change, https://github.com/commercetools/commercetools-adyen-integration/pull/232#discussion_r404345695

extension/test/integration/get-origin-keys.handler.spec.js
ahmetoz: I removed those tests, This is the same change, #232 (comment)

extension/src/config/config.js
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/bc45046b67a4fdb6aae6f4e0512a25470d0e5607

extension/src/paymentHandler/payment-utils.js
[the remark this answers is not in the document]
ahmetoz: I will add this as a todo, and we need to add unit tests for the payment-utils but this parts could be changed later, also we don't expect a Localised Enum field on the response custom fields.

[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/a8ca5a2000d3db9df9d0987f580c9103a59d5eb9

extension/test/integration/get-origin-keys.handler.spec.js
ahmetoz: it seems we have some issues created for error handling: https://github.com/commercetools/commercetools-adyen-integration/issues/46

extension/src/config/config.js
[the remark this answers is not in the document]
ahmetoz: related issue: https://github.com/commercetools/commercetools-adyen-integration/issues/173

extension/src/paymentHandler/payment-utils.js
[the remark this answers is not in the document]
ahmetoz: related issue: https://github.com/commercetools/commercetools-adyen-integration/issues/75#issuecomment-514196727

extension/src/paymentHandler/payment-handler.js
```diff
@@ -1,57 +1,42 @@
...
-  const handler = _getPaymentHandler(paymentObject)
-  const handlerResponse = await handler.handlePayment(paymentObject)
-  if (handlerResponse.errors)
-    return { success: false, data: handlerResponse }
+
+  const handlers = _getPaymentHandlers(paymentObject)
+  const handlerResponses = await Promise.all(
+    handlers.map(handler => handler.handlePayment(paymentObject)))
+  const handlerResponse = {
+    actions: handlerResponses.flatMap(result => result.actions)
```
butenkor, line 28: Can it happen that one or more responses will have no actions defined? If yes than flatMap will generate an `undefined` action element into the actions array.

extension/src/paymentHandler/get-origin-keys.handler.js
```diff
@@ -0,0 +1,18 @@
+const pU = require('./payment-utils')
+const c = require('../config/constants')
+const { getOriginKeys } = require('./web-component-service')
+
+async function handlePayment (paymentObject) {
```
butenkor, line 5: Maybe instead of `handlePayment` we can call it `execute` or just `handle` as it conflicts a bit with this: https://github.com/commercetools/commercetools-adyen-integration/pull/232/files#diff-b7da223fe1867c89db8963c55099bdccR17

extension/src/paymentHandler/payment-utils.js
```diff
@@ -59,25 +59,25 @@ function getMatchingCtpState (adyenState) {
...
       response: JSON.stringify(response),
-      request: JSON.stringify(request.body),
-      type,
-      status
+      request: request.body,
```
butenkor, line 74: You do not need to escape the body JSON or it will be escaped automatically?

extension/src/paymentHandler/payment-utils.js
```diff
@@ -104,11 +104,12 @@ function createChangeTransactionStateAction (transactionId, transactionState) {
   }
 }
 
-function createSetCustomFieldAction (name, value) {
+function createSetCustomFieldAction (name, response) {
   return {
     action: 'setCustomField',
     name,
-    value
+    // todo (ahmet): Don't stringify here. Sometimes you want to pass JSON object, e.g. for LocalizedEnum.
```
butenkor, line 111: I have seen a couple of TODOs in this PR. Why not addressing them right away?

extension/src/paymentHandler/payment-utils.js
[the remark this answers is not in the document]
butenkor: @lojzatran Is there any case in the web-components where we need to persist anything else than just JSON?

extension/src/paymentHandler/web-component-service.js
```diff
@@ -0,0 +1,50 @@
+const fetch = require('node-fetch')
```
butenkor, line 1: > extension/src/paymentHandler/web-component-service.js

I wonder why this service is nested within paymentHandler package. It is not really a handler, right?

extension/src/paymentHandler/web-component-service.js
```diff
@@ -0,0 +1,50 @@
...
+    response = await fetchAsync(endpoint, request)
+  } catch (err) {
+    response = serializeError(err)
+  }
+  return { request, response }
+}
+
+function buildRequest (requestString) {
+  const body = JSON.stringify({
+    merchantAccount: config.adyen.merchantAccount,
```
butenkor, line 28: What if merchantAccount is already supplied within payload?

extension/src/paymentHandler/web-component-service.js
```diff
@@ -0,0 +1,50 @@
...
+    ...JSON.parse(requestString)
+  })
+
+  return {
+    method: 'POST',
+    body: body,
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Api-Key': config.adyen.apiKey
```
butenkor, line 37: Is it removed before we persist it as interface interaction?

extension/src/paymentHandler/get-origin-keys.handler.js
ahmetoz: you're absolutely right!

extension/src/paymentHandler/web-component-service.js
ahmetoz: You're right.

extension/src/paymentHandler/web-component-service.js
ahmetoz: This is a good example, and yes possible, I've also a scenario that the requestString would be not a valid JSON, let me think what I could design a better structure for those... I could not find a better check yet, and this is a todo on a test. I will try to address them.

extension/src/paymentHandler/web-component-service.js
ahmetoz: yes, it is. but that part is not clear from our implementations (was not clear before too), maybe I should improve the design of creating actions both for interface interaction and custom fields.

extension/src/paymentHandler/payment-utils.js
ahmetoz: you don't but the same concern as here: https://github.com/commercetools/commercetools-adyen-integration/pull/232#discussion_r405636607. I will try to change those to find a better structure.

extension/src/paymentHandler/payment-handler.js
ahmetoz: You're right, error handling and validations for edge cases are not well structured (which I added as a todo). Let me check what I could do for those.

extension/src/paymentHandler/payment-utils.js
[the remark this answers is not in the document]
lojzatran: So far no.

lojzatran:
Can you check eslint and reformat your code?

ahmetoz:
> Can you check eslint and reformat your code?

addressed : https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/599533d47cfcbc305f5e47c391561999c531f318

extension/src/config/constants.js
```diff
@@ -1,7 +1,9 @@
 module.exports = {
   SUCCESS: 'SUCCESS',
   FAILURE: 'FAILURE',
-  CTP_INTERFACE_INTERACTION: 'ctp-adyen-integration-interaction',
+  CTP_ADYEN_INTEGRATION: 'ctp-adyen-integration',
+  CTP_INTERFACE_INTERACTION_PAYMENT_TYPE_KEY: 'ctp-adyen-integration-interaction-payment-type',
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/49ffdfed4070b1dc2405e4e9774bb92ab986b9e8

extension/src/paymentHandler/web-component-service.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/6e68777b0d3e0f952691b8bdfa46b84ceafd6e20

extension/src/paymentHandler/get-origin-keys.handler.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/08c0812356cab3fb7d309798b7a5727c9fb204fa

extension/src/paymentHandler/payment-utils.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/6777ff34a8005fa3b01970789ff2146018efc635

extension/src/paymentHandler/payment-utils.js
ahmetoz: I removed that todo, as all request and responses will be passed as an object to those utilities. https://github.com/commercetools/commercetools-adyen-integration/pull/232#discussion_r406259171

extension/src/paymentHandler/web-component-service.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/blob/231-add-payment-methods-and-origin-keys/extension/src/web-component-service.js#L34-L39

ahmetoz: additionally removing additional data from the response is moved here from payment-utils.

extension/src/paymentHandler/payment-handler.js
ahmetoz: For now, I would not expect any kind of unexpected error into the execute method, for the invalid JSON cases I added it as validation before payment handler executions:  https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/29805e1c742c6ae3cfa0cee5832ee9f247a32527

extension/src/validator/validator-builder.js
```diff
@@ -146,4 +129,17 @@ function withPayment (paymentObject) {
   }
 }
 
+function invalidJSON (requestString) {
```
lojzatran, line 132: It's quite hard to read negative function. Better is to have `isValidJson()`.

extension/src/paymentHandler/payment-handler.js
```diff
@@ -1,57 +1,42 @@
...
-    return paymentHandlers.kcpHandler
-  return paymentHandlers.fetchPaymentMethodsHandler
-}
+function _getPaymentHandlers (paymentObject) {
+  // custom field on payment is not a mandatory field.
+  if (!paymentObject.custom)
+    return []
 
+  const handlers = []
+  if (paymentObject.custom.fields.getOriginKeysRequest && !paymentObject.custom.fields.getOriginKeysResponse)
```
lojzatran, line 35: I guess we will have soon more of these if clauses. If you want, you can put these condition checkings in the handlers, put the handlers in an array and here just iterate through this array and call the checking methods. Sth like this:
```
const availableHandlers = [getPaymentMethodsHandler, getOriginKeysHandler]
const handlers = []
for (const handler of handlers)
    if (handler.isSuitable(paymentObject))
        handlers.push(handler)
return handlers
```
But maybe this is not so readable. You can decide.

extension/src/web-component-service.js
```diff
@@ -0,0 +1,52 @@
+const fetch = require('node-fetch')
```
lojzatran, line 1: You can put this class into `utils` or `service` folder.

extension/src/validator/validator-builder.js
ahmetoz: but then you will check it with an `!`

extension/src/web-component-service.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+  return { request: requestObj, response }
+}
+
+async function fetchAsync (endpoint, requestObj) {
+  const request = buildRequest(requestObj)
+  const response = await fetch(`${config.adyen.apiBaseUrl}/${endpoint}`, request)
+  const responseBody = await response.json()
+  // strip away sensitive data from the adyen response.
+  delete responseBody.additionalData
```
butenkor, line 30: I wonder if we might have also sensitive data in our payload requests which we both save as payment custom fields and interface interactions?

extension/src/web-component-service.js
ahmetoz: IMO putting some into folders is not helpful if you have a small-sized project like extension module, I did not think current structure is also right, I am considering we will change those structures a little bit later.

lojzatran: OK let's do this change later when we have more files, so we know how to structure better.

extension/src/paymentHandler/payment-handler.js
ahmetoz: the structure looks ok, deciding which handler will work is not depend only on the handler, because handlers have side effects, let's say checking makePayment and submitPaymentDetails should not be work parallel together, so in makePayment handler checking submitPaymentDetail request will not be a good practice, probably validation should be outside the handler.

extension/src/paymentHandler/web-component-service.js
butenkor: Where is it removed? I would expect that there is a single place where both request and response is stripped from sensitive data.

extension/src/paymentHandler/payment-handler.js
lojzatran: OK let's wait and refactor later then.

extension/src/paymentHandler/web-component-service.js
ahmetoz: I am not removing it because the data is stored as a header, persisting only the body is enough.

extension/src/web-component-service.js
ahmetoz: I am not %100 sure what kind of data is sensitive. We analyzed it before and come up only for that part is sensitive, but yes we should keep in mind we would need to strip different fields later.

extension/src/paymentHandler/web-component-service.js
ahmetoz: >I would expect that there is a single place where both request and response is stripped from sensitive data.

I moved those parts from utils to web-components-service file, so it's now in one place, also all adyen calls are in one place.

extension/src/validator/validator-builder.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/232/commits/19456341b3d610e4999a990dafc5cc7f19f7a548

Issue #277, opened 2020-05-12
Verify amountPlanned changes


lojzatran:
Amount planned has to be set by the user on payment creation. amountPlanned can be changed any time as long as no makePayment request has been executed yet. In case makePayment is executed make sure that amountPlanned is always the same as the amount within makePayment interface interaction request. In case they are different, discard payment update request

PR #283, opened 2020-05-13
#277 verify amountPlanned

lojzatran:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/277


extension/src/paymentHandler/payment-handler.js
```diff
@@ -12,7 +12,9 @@ async function handlePayment (paymentObject) {
     // if it's not adyen payment, ignore the payment
```
ahmetoz, line 12: not related to this branch, but could you remove lodash dependency from that file if we are not using it.
` error  '_' is assigned a value but never used`

extension/src/validator/error-messages.js
```diff
@@ -15,5 +15,6 @@ module.exports = {
   GET_PAYMENT_METHODS_REQUEST_INVALID_JSON: 'getOriginKeysRequest does not contain valid JSON.',
   MAKE_PAYMENT_REQUEST_INVALID_JSON: 'makePaymentRequest does not contain valid JSON.',
   // eslint-disable-next-line max-len
-  SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON: 'submitAdditionalPaymentDetailsRequest does not contain valid JSON.'
+  SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON: 'submitAdditionalPaymentDetailsRequest does not contain valid JSON.',
+  AMOUNT_PLANNED_INVALID: 'amountPlanned field cannot be changed after makePayment request is made to Adyen'
```
ahmetoz, line 19: it's valid but immutable, maybe change the naming in here to 
"AMOUNT_PLANNED_MODIFICATION_IS_NOT_ALLOWED"

Also, the message here is for the user, could we say more precisely: 

`Modifying the "amountPlanned" value is not allowed after the payment amount charged by Adyen.`

extension/src/validator/validator-builder.js
```diff
@@ -20,6 +21,23 @@ function withPayment (paymentObject) {
           = errorMessages.SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON
       return this
     },
+    validateAmountPlanned () {
```
ahmetoz, line 24: The function is not tested directly, but indirectly covered in payment-handler tests, why?

extension/src/validator/validator-builder.js
```diff
@@ -20,6 +21,23 @@ function withPayment (paymentObject) {
...
+      const oldMakePaymentRequestObj = paymentObject.interfaceInteractions
```
ahmetoz, line 25: is the oldest or newest ?

lojzatran: It's the newest from the old ones 😄

extension/src/validator/validator-builder.js
lojzatran: Why not? https://github.com/commercetools/commercetools-adyen-integration/pull/283/commits/b30637e690c0d66bfd4f3ba43523bb090a5fd507

extension/src/validator/error-messages.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/283/commits/fd30c79d4afcf8cbb87739bb7a3d268700fc9e51

I will not change the message to your message as it's probably not true - amountPlanned is not charged in some cases, etc. you need to do the redirect and then verification etc. 
And the real reason for this error is simply after there's a makePayment request, amountPlanned cannot be changed.

extension/src/paymentHandler/payment-handler.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/283/commits/ddde5580e2e422d302a20bc0195f3a9149b58a1f

PR #285, opened 2020-05-14
Update payment-interface-interaction-types.json

peniakoff:
We can not create array of custom types as a custom type in the Commerce Tools. That is reason I suggest remove curly braces for that JSON object.


butenkor:
Hi @peniakoff. I wonder why you create those types manually as this file was rather meant to be used here: https://github.com/commercetools/commercetools-adyen-integration/blob/4811e04e7f804dfcf2efb2a7ad7b215e738d9e83/extension/src/config/init/ensure-interface-interaction-custom-type.js#L9

peniakoff:
Hi, @butenkor , we are creating all requires custom types by CT in different initialization process - no by this JS code. That's why I was wonder too - custom type can not be created because of array.

ahmetoz:
Hey @peniakoff ✋ 

> we are creating all requires custom types by CT in different initialization process

I am not sure if I understand your scenario. Could you please give me more details?

BTW if you want to create custom type manually you could use the option (environment variable): `ENSURE_RESOURCES=false`, so integration will not create resources, which I don't suggest if you don't have a really valid reason.

You could also check it in the [documentation](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/DevelopmentAndDeploymentGuide.md#run-module) as :
> Set to false to disable the creation of resources in commercetools (e.g. custom types)

peniakoff:
I mean we have Java based app to initializing all needed custom types (via Java SDK and Custom Types based on JSONs). We are using there pure JSONs to upload new types. There's also _payment-interface-interaction-types.json_ which we can use only as one object - no array of types. Maybe the name of the json file is is misleading? Look, _payment-custom-types.json_ contains in the name also the plural of types, but inside the JSON is only one custom type. The same situation should be regarding the _payment-interface-interaction-types.json_, nope?

butenkor:
True `payment-interface-interaction-types.json` does not have to be an array. I guess in the beginning we considered using multiple types for each payment method and this approach would allow adding new types without changing the code.

As linked your PR would break the existing code. Please change the code accordingly.

peniakoff:
Hii, @butenkor , PR was updated right now. Please, check it and feel free to comment. Thanks!

ahmetoz:
@peniakoff https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v4.0.1 the change is released, thanks!

peniakoff:
Thank you too, @ahmetoz, for your cooperation! :blush:

Issue #282, opened 2020-05-13
Adapt refund/cancel transaction to web-components flow.


ahmetoz:
Backend (not my payments/front end) should be able to add a refund transaction with state "Initial" and an extension module should be able to process it with Adyen automatically and map the response back to the refund transaction state.

PR #286, opened 2020-05-18
Adapt refund/cancel transaction to web-components flow. #282

ahmetoz:
Issue: #282


extension/src/paymentHandler/cancel-or-refund.handler.js
```diff
@@ -1,51 +1,63 @@
...
-    body: JSON.stringify(body),
-    headers: { 'x-api-key': config.adyen.apiKey, 'Content-Type': 'application/json' }
+  const interfaceInteraction = paymentObject.interfaceInteractions
+    .find(interaction => interaction.fields.type === CTP_INTERACTION_TYPE_MAKE_PAYMENT)
+  if (interfaceInteraction && !isEmpty(interfaceInteraction.fields.response)) {
+    // OPTIONAL:
+    // Specifies a unique identifier for payment modification.
+    // The reference field is useful to tag a partial cancel or refund for future reconciliation.
+    const makePaymentResponse = JSON.parse(interfaceInteraction.fields.response)
+    cancelOrRefundRequestObj.reference = makePaymentResponse.merchantReference
```
lojzatran, line 42: Why do you put `merchantReference` for this field? There's nothing about this part in the documentation. Where did you find it?

extension/test/integration/cancel-or-refund.handler.spec.js
```diff
@@ -1,62 +1,91 @@
...
-          transaction: {
-            type: 'Refund',
-            amount: {
-              currencyCode: transaction.amount.currencyCode,
-              centAmount: transaction.amount.centAmount
-            },
-            state: 'Initial'
-          }
+  it('given a payment ' +
+    'when a transaction with type "Refund" with "Initial" state, ' +
```
lojzatran, line 61:
```suggestion
    'when a transaction with type "Refund" with "Initial" state exists, ' +
```

extension/test/unit/cancel-or-refund.handler.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      id: 'transaction2',
+      type: 'Refund',
+      state: 'Initial'
+    })
+
+    const response = await execute(paymentObject)
+    assertUpdateActions(response)
+  })
+
+  function assertUpdateActions ({ actions }) {
```
lojzatran, line 62: Can you put this helper method after all the tests?

extension/test/integration/cancel-or-refund.handler.spec.js
```diff
@@ -1,62 +1,91 @@
...
+      paymentMethod: {
+        type: 'scheme',
+        encryptedCardNumber: 'test_4111111111111111',
+        encryptedExpiryMonth: 'test_03',
+        encryptedExpiryYear: 'test_2030',
+        encryptedSecurityCode: 'test_737'
+      },
+      returnUrl: 'https://your-company.com/'
+    }
+    const paymentDraft = {
```
lojzatran, line 33: Is there an `authorizationTransaction` in this payment? I saw that you later use this transaction to get `originalReference` and according to the docs, it's a must have param.

extension/src/paymentHandler/cancel-or-refund.handler.js
ahmetoz: General documentation is here: https://docs.adyen.com/checkout/cancel-or-refund#cancel-or-refund-a-payment 

a reference that is passed with /payment request: https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v52/post/payments__reqParam_reference 
will return in response as a merchant reference.

Check here for an example: https://github.com/commercetools/commercetools-adyen-integration/blob/01d482cc780f0bc23bbc5cd0ecc58abade49d5c1/extension/test/unit/cancel-or-refund.handler.spec.js#L29-L32

As described in the comments, it's an optional field. Documentations do not say merchant reference because it could be any reference, so merchant reference could be a reference field for our scenario. Before web-components that field was a custom field passed by the front end. Now we are using the `makePayment` interaction to get the reference.

After that long comment, if you're asking a different thing, please give an example of your question.

extension/test/integration/cancel-or-refund.handler.spec.js
ahmetoz: No, it will be created by integration, 2 phases. So this is before cancel or refund, that way integration test will make a payment first _**before**_ the test and then payment could have an originalReference(pspReference).

extension/test/unit/cancel-or-refund.handler.spec.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/286/commits/01d482cc780f0bc23bbc5cd0ecc58abade49d5c1

extension/src/paymentHandler/cancel-or-refund.handler.js
lojzatran: No, that's what I was asking for :) Theoretically there could be multiple `interfaceInteraction` for `makePayment`, so I guess you should get the newest one. Is it better if you take the response from the custom field `makePaymentResponse`?

ahmetoz: > Theoretically there could be multiple interfaceInteraction for makePayment, so I guess you should get the newest one.

This is a good point 👍

ahmetoz: The issue is created for the additional payment details : 
https://github.com/commercetools/commercetools-adyen-integration/issues/289

Refactoring for the comment: https://github.com/commercetools/commercetools-adyen-integration/pull/286/commits/d0baa3b9256518b24b8ced13c973d8e3c099f24a

Issue #288, opened 2020-05-22
Support Klarna


lojzatran:
Add logic to add line item and custom line item information for payments like Klarna and make sure that we do not override existing line items (flexibility for the front end). See an example here: https://github.com/commercetools/commercetools-payone-integration/blob/d05d56ee296549b19119fdffcf338131c74ed0f1/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequestWithCart.java#L178

PR #294, opened 2020-05-29
#288 klarna

lojzatran:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/288


extension/src/config/klarna-constants.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  PAYMENT_METHOD_TYPE_KLARNA_METHODS: ['klarna', 'klarna_paynow', 'klarna_account'],
+  ADYEN_PERCENTAGE_MINOR_UNIT: 10000,
+  DEFAULT_PAYMENT_LANGUAGE: 'en',
```
ahmetoz, line 4: Why default is en ?

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function _getShippingMethodDescription (shippingInfo) {
+  const shippingMethod = shippingInfo.shippingMethod.obj
+  if (shippingMethod)
+    return shippingMethod.description
+  return shippingInfo.shippingMethodName
+}
+
+function _localizeOrFallback (localizedString,
```
ahmetoz, line 74: You could skip that comment:

I guess, fallback does not mean this, fallback is more like a method 🤔 
Here I would use something like `_getLocalisedValueOrElse(localisedString, locales, defaultValue)`

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+  return locales
+}
+
+function _createAdyenLineItemFromLineItem (ctpLineItem, locales) {
+  return {
+    id: ctpLineItem.variant.sku,
+    quantity: ctpLineItem.quantity,
+    description: _localizeOrFallback(ctpLineItem.name, locales, KLARNA_DEFAULT_LINE_ITEM_NAME),
+    amountIncludingTax: ctpLineItem.price.value.centAmount,
+    taxPercentage: ctpLineItem.taxRate.amount * ADYEN_PERCENTAGE_MINOR_UNIT
```
ahmetoz, line 43: taxRate is an optional field, do we need to do a check here to avoid NPEs?

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+  return {
+    id: ctpLineItem.id,
+    quantity: ctpLineItem.quantity,
+    description: _localizeOrFallback(ctpLineItem.name, locales, KLARNA_DEFAULT_LINE_ITEM_NAME),
+    amountIncludingTax: ctpLineItem.money.centAmount,
+    taxPercentage: ctpLineItem.taxRate.amount * ADYEN_PERCENTAGE_MINOR_UNIT
+  }
+}
+
+function _createShippingInfoAdyenLineItem (shippingInfo) {
```
ahmetoz, line 57: shipping info is an optional field in the cart,  do we need to do a check here to avoid NPEs?
OK... I could find different places to find checks, here what's the strategy for payment handler, if the cart does not have enough info for lineItem mapping for klarna ?

extension/src/paymentHandler/make-payment.handler.js
```diff
@@ -26,4 +35,16 @@ async function execute (paymentObject) {
   }
 }
 
+async function _fetchMatchingCart (paymentObject) {
```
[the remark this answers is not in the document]
lojzatran: It does not do anything Klarna specific, so I didn't want to put it there.

extension/src/config/klarna-constants.js
lojzatran: Don't know why, but it's according to payone: https://github.com/commercetools/commercetools-payone-integration/blob/dfb14f4f17c356fa0d14b7f007aebe76065942ec/service/src/main/java/com/commercetools/util/ServiceConstants.java#L10

extension/src/service/web-component-service.js
```diff
@@ -1,6 +1,6 @@
 const fetch = require('node-fetch')
```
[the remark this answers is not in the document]
lojzatran: Benefit is that you have it organized :) We can put them all in one folder, but then it's lots of files at once. That's the idea of packages in java as well. There are several ways to group the files into packages, and in our case, I see grouping them by type as most suitable.

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
+const {
```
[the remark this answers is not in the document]
lojzatran: I think it's a service, why not? But it can also be an utility or part of the klarna handler as well, I think it's kind of the same here. WDYT?

What do you mean that `createLineItems` should be encapsulated? Isn't it already encapsulated? It's the only method that is exposed in this service.

extension/src/config/klarna-constants.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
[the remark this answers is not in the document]
lojzatran: Only in tests.

extension/src/paymentHandler/make-payment.handler.js
```diff
@@ -1,9 +1,18 @@
...
+const { makePayment } = require('../service/web-component-service')
+const { createLineItems } = require('../service/klarna-service')
 const pU = require('./payment-utils')
 const c = require('../config/constants')
+const { PAYMENT_METHOD_TYPE_KLARNA_METHODS } = require('../config/klarna-constants')
+const ctpClientBuilder = require('../ctp/ctp-client')
 
 async function execute (paymentObject) {
   const makePaymentRequestObj = JSON.parse(paymentObject.custom.fields.makePaymentRequest)
+  if (_isKlarna(makePaymentRequestObj) && !makePaymentRequestObj.lineItems) {
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/36769017ff39cb116c9c24cc01a50db02ce4ceef

extension/src/service/klarna-service.js
lojzatran: Hm good point. I had a look in the docs and it says this:
```
Will be set automatically in the Platform TaxMode once the shipping address is set is set. For the External tax mode the tax rate has to be set explicitly with the ExternalTaxRateDraft.
```
So the question is does it make sense to check it here? When you make a payment, should you already have a cart with the shipping address already? Otherwise I would not make sense, right? Theoretically there could be NPE here, but practically I don't see a value for check. Maybe @butenkor can clarify here.

extension/src/service/klarna-service.js
lojzatran: Same idea as above. In the docs it says: `Set automatically once the ShippingMethod is set.`. Once you make a payment, I would expect ShippingMethod to be set.

extension/src/service/klarna-service.js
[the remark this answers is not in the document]
ahmetoz: It's not a service because the logic here is **only** part of the internals of the klarna payment handling (execute method) and it's just doing line items construction, you could rename it maybe sth like line-item-mapper, but it should not be part of the current scope when we got this as a new requirement we could extract those to not duplicate same logic but current scope does not need it.

In the project, we have only one service pattern which is web-component-service, and it's a service because this file is organized for reusing Adyen communication and it does not depend on any other execute method, even for that purpose we are not directly testing it because it could be changed easily.

Actually main reason is not naming, in detail here for that logics there is no reusability purpose, naming this as a service, exposing it and directly testing this instead of testing payment-handler: execute method is wrong. 
 write a test for your contracts not for your internal architectural models.

extension/src/config/klarna-constants.js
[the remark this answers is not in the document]
ahmetoz: I don't follow the code-sharing in here if it's not used anywhere in any other parts of the code, those are constants but should not be exposed, please define your constants as closest as possible.

extension/src/service/klarna-service.js
ahmetoz: If it's not possible you don't need to check it if it's possible it will block the process and the error will not be stored in interface interaction. Also how you will be sure that a cart has all the information that you need?

extension/src/service/klarna-service.js
ahmetoz: It's not hard to understand, it's a review comment to make it better, you could skip.

extension/src/config/klarna-constants.js
ahmetoz: Please clarify this, maybe it should not be default but mandatory.

extension/src/service/klarna-service.js
butenkor: You would need to try it out but according to docs, you can have a cart without shipping address but with `totalPrice` set (and no tax info on cart and line item level). Checking for NPE might be a good idea here.

extension/src/service/web-component-service.js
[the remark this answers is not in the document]
ahmetoz: Here I will try to put my ideas so it's just matter of taste and also it's not a big issue to separate them by types, but also you need to consider that the extension module is a microservice, it could make sense to separate them if you have lots of files in a folder but it's not a case here, IMO it's layering the implementation unnecessarily **outside** of the payment handler package is wrong, which is based on the file namings, the file name already has this additional layer information so it could be in the same package or could be inside a sub-package called service but **not outside**. I guess for the web-components-service it was also in the wrong package before 🤦 Anyways we could go back that package organization inside the extension module later separately.

IMO in java packages idea is not that way like grouping by types, for example, in java SDK packages grouping is not based on the internal layers, it's rather separated by development functionality, for example, take `java. util.concurrent` package into consideration, the package groups the concurrency utilities that you could use, so in that package, most of the things are in one place if it does have lots of different use cases or code-sharing only then java packages are separated inside that package for example:  `java.util.concurrent.atomic`, so the java package groupings are not putting additional folders like `java.util.concurrent.future` because Future is used all in the same level of the `java.util.concurrent` package.

extension/src/paymentHandler/make-payment.handler.js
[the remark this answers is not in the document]
ahmetoz: no, It's only klarna specific, fetching cart only needed for klarna make payment. I will resolve this because it's resolved in a different commit: 3676901

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+
+  lineItems.push(_createShippingInfoAdyenLineItem(cart.shippingInfo))
+
+  return lineItems
+}
+
+function _getLocales (cart, payment) {
+  const locales = []
+  let paymentLanguage = payment.custom && payment.custom.fields['languageCode']
+  if (!paymentLanguage)
```
ahmetoz, line 27: refactoring point: these checks are not easy to reason, what happens to if-else cases or switch cases?

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+  const shippingMethod = shippingInfo.shippingMethod.obj
+  if (shippingMethod)
+    return shippingMethod.description
+  return shippingInfo.shippingMethodName
+}
+
+function _localizeOrFallback (localizedString,
+                              locales,
+                              fallback) {
+  let value
```
ahmetoz, line 77: refactor point: maybe write the code filtering code with the functional way (maybe sth like find) instead of imperative style ?

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+    id: `${shippingInfo.shippingMethodName}`,
+    quantity: 1,
+    description: _getShippingMethodDescription(shippingInfo),
+    amountIncludingTax: shippingInfo.price.centAmount,
+    taxPercentage: shippingInfo.taxRate.amount * ADYEN_PERCENTAGE_MINOR_UNIT
+  }
+}
+
+function _getShippingMethodDescription (shippingInfo) {
+  const shippingMethod = shippingInfo.shippingMethod.obj
```
ahmetoz, line 68: refactor point: this could be written as a one-liner function.

extension/src/service/klarna-service.js
```diff
@@ -0,0 +1,86 @@
...
+    description: _localizeOrFallback(ctpLineItem.name, locales, KLARNA_DEFAULT_LINE_ITEM_NAME),
+    amountIncludingTax: ctpLineItem.money.centAmount,
+    taxPercentage: ctpLineItem.taxRate.amount * ADYEN_PERCENTAGE_MINOR_UNIT
+  }
+}
+
+function _createShippingInfoAdyenLineItem (shippingInfo) {
+  return {
+    id: `${shippingInfo.shippingMethodName}`,
+    quantity: 1,
```
ahmetoz, line 60: why it's 1?

extension/test/unit/make-payment.handler.spec.js
```diff
@@ -227,4 +229,89 @@ describe('make-payment::execute', () => {
     expect(addTransaction.transaction.state).to.equal('Failure')
     expect(addTransaction.transaction.interactionId).to.equal(JSON.parse(paymentErrorResponse).pspReference)
   })
+
```
ahmetoz, line 232: I guess it's better if you separate those tests into a different file.

extension/test/unit/service/klarna-service.spec.js
```diff
@@ -0,0 +1,98 @@
+const { expect } = require('chai')
+const _ = require('lodash')
+const klarnaService = require('../../../src/service/klarna-service')
+const { KLARNA_DEFAULT_LINE_ITEM_NAME, DEFAULT_PAYMENT_LANGUAGE } = require('../../../src/config/klarna-constants')
+const ctpCart = require('../fixtures/ctp-cart')
+
+describe('klarna-service::createLineItems', () => {
```
ahmetoz, line 7: I guess those tests should be part of the make-payment unit tests.

extension/src/service/klarna-service.js
lojzatran: I also tried to refactor this method, but I don't know how. If you have a suggestion, please write it here :)

extension/src/service/klarna-service.js
lojzatran: Do you mean `if` condition below? I think it doesn't make much difference, so I will leave it.

extension/src/service/klarna-service.js
lojzatran: I don't know how to refactor 🤔 Can you give your suggestion?

extension/src/config/klarna-constants.js
lojzatran: It's a language field that should be used for getting the name for a product. The shop can set the language on the payment or on the cart. If it's not set, it will pick a default language, which is 'en' in this case. If even 'en' is not there, it will just send "ITEM". I don't know why it's 'en' and not 'de' or 'fr' or some other language code, but I think 'en' is quite reasonable. It cannot be mandatory as the shop can choose any language they want. https://github.com/commercetools/commercetools-adyen-integration/pull/294/files#diff-1519468bcea5d71c0468f5405f237e61R24

extension/src/service/klarna-service.js
lojzatran: Because you can have only 1 shipping per cart.

lojzatran: See more: https://github.com/commercetools/commercetools-payone-integration/blob/dfb14f4f17c356fa0d14b7f007aebe76065942ec/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequestWithCart.java#L290

extension/src/service/web-component-service.js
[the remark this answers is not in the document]
lojzatran: You're right that the grouping by type it's not very good, but I don't see a better way here. So do you suggest to put everything under `src` folder? Isn't it too many files in one folder if we mix `paymentHandlers`, `validators`, `service` etc.?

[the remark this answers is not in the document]
ahmetoz: I will skip this comment if you could not see the better way, also according to my other comments there is no klarna service, so the package will only include web-components-service.

extension/src/service/klarna-service.js
ahmetoz: I don't want to spend time on this, but the review is obvious I could not understand that code, maybe sth like below (I did not check the tests)

```
if (paymentLanguage)
    locales.push(paymentLanguage)
  else
    paymentLanguage = cart.locale
    if (paymentLanguage !== DEFAULT_PAYMENT_LANGUAGE)
      locales.push(paymentLanguage)
```

extension/src/service/klarna-service.js
ahmetoz: This is not clear when you read the code, please add this as a comment.

extension/src/service/klarna-service.js
ahmetoz: Did you try this: locales.find(element => localisedString[element]) ? Even it could be better with additional map.

```
const locale = locales.find(locale => localizedString[locale])
return locale ? localizedString[locale] : fallback
```

extension/src/config/klarna-constants.js
ahmetoz: could you please add that decision as a comment, otherwise it's hard to reason.

extension/src/service/klarna-service.js
lojzatran: Yes, it's not easy to understand. This fallback algorithm is complex in [payone](https://github.com/commercetools/commercetools-payone-integration/blob/dfb14f4f17c356fa0d14b7f007aebe76065942ec/service/src/main/java/com/commercetools/pspadapter/payone/domain/payone/model/common/AuthorizationRequestWithCart.java#L183) as well. Here I already tried to simplify it. Your suggestion is changing the method behaviour. I added a comment to make it more clear: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/419c4a50172ef8f2d667dc57bc5fd28eaadf0703#diff-4d2f466e11952575f6b8048a0f137b16R46-R50

extension/src/config/klarna-constants.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/419c4a50172ef8f2d667dc57bc5fd28eaadf0703#diff-4d2f466e11952575f6b8048a0f137b16R68-R72

extension/src/service/klarna-service.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/419c4a50172ef8f2d667dc57bc5fd28eaadf0703#diff-4d2f466e11952575f6b8048a0f137b16R97

extension/test/unit/service/klarna-service.spec.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/2124321c5cdf820ced051c2882c217d9f45c1d2c

extension/src/config/klarna-constants.js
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/25e2649c5cde632f802acff47324c197c41f0277

extension/src/service/klarna-service.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/9e1dda9c44de2925f748322e60260b8b15f95fa0

lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/77a5ed9c4c2351520d5209402b927fb9b6967632

extension/src/service/klarna-service.js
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/143cc441a51b4debc177f6f04f19c8a1a99b90c1

extension/src/service/klarna-service.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/e1fb9e8fec635a465b3b350798818a0c3279bee8

extension/src/service/klarna-service.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/294/commits/e1fb9e8fec635a465b3b350798818a0c3279bee8

extension/test/unit/make-payment.handler.spec.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/commit/2124321c5cdf820ced051c2882c217d9f45c1d2c

PR #297, opened 2020-06-01
extension/src/lambda.js - handle nulls correctly

jamiegardiner:
Null check around `paymentResult.data`, which can be null if `paymentHandler` short circuits when not an adyen payment

### DIFF
--- a/extension/src/lambda.js
+++ b/extension/src/lambda.js
@@ -9,8 +9,8 @@ exports.handler = async function (event) {
 
     return {
       responseType: paymentResult.success ? 'UpdateRequest' : 'FailedValidation',
-      errors: paymentResult.data.errors,
-      actions: paymentResult.data.actions
+      errors: paymentResult.data ? paymentResult.data.errors : undefined,
+      actions: paymentResult.data ? paymentResult.data.actions : undefined
     }
   } catch (e) {
     logger.error(e, `Unexpected error when processing event ${JSON.stringify(event)}`)
--- a/extension/test/unit/lambda.spec.js
+++ b/extension/test/unit/lambda.spec.js
@@ -40,6 +40,16 @@ describe('Lambda handler', () => {
     expect(result.actions).to.equal(undefined)
   })
 
+  it('does not throw unhandled exception when handlePayment data is null', async () => {
+    sinon.stub(paymentHandler, 'handlePayment').returns({ success: true, data: null })
+
+    const result = await handler(event)
+
+    expect(result.responseType).equals('UpdateRequest')
+    expect(result.errors).equals(undefined)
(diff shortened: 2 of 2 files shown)


ahmetoz:
@jamiegardiner Thanks for your first contribution 🎉  
We will try to check and review in a couple of days 🙏

extension/test/unit/lambda.spec.js
```diff
@@ -40,6 +40,16 @@ describe('Lambda handler', () => {
     expect(result.actions).to.equal(undefined)
   })
 
+  it('does not throw unhandled exception when handlePayment data is null', async () => {
+    sinon.stub(paymentHandler, 'handlePayment').returns({ success: true, data: null })
+
+    const result = await handler(event)
+
+    expect(result.responseType).equals('UpdateRequest')
```
butenkor, line 48: It seems that in the described case it should rather say `RequestIgnored` as there will be no updates applied?

jamiegardiner: Hi @butenkor, as this will be running in a AWS Lambda the only valid `responseType` that commercetools will allow is either `UpdateRequest` or `FailedValidation`. We don't want failed validation in this case because this will cause all updates in the transaction to rollback.

https://docs.commercetools.com/http-api-projects-api-extensions#response

Also in this case we don't know the request is actually ignored, we only know paymentHandler returned a `{success: true, data: null}`, the change is to just handle the case of `{ data: null }`, which currently causes the code to bomb out

butenkor: Agree

extension/src/lambda.js
```diff
@@ -9,8 +9,8 @@ exports.handler = async function (event) {
 
     return {
       responseType: paymentResult.success ? 'UpdateRequest' : 'FailedValidation',
-      errors: paymentResult.data.errors,
-      actions: paymentResult.data.actions
+      errors: paymentResult.data ? paymentResult.data.errors : undefined,
+      actions: paymentResult.data ? paymentResult.data.actions : undefined
```
butenkor, line 13: Could you please add a comment to the code from your PR description as it is not obvious to the reader:
`Null check around paymentResult.data, which can be null if paymentHandler short circuits when not an adyen payment`

ahmetoz:
My mistake 🤦  @butenkor already approved it, so I merged it, I will add the comment to the code. 
https://github.com/commercetools/commercetools-adyen-integration/pull/297#discussion_r434463897

ahmetoz:
@jamiegardiner https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v4.0.1 the fix is released, thanks!

PR #301, opened 2020-06-03
fix: return empty actions when paymentHandler short-circuits

jamiegardiner:
The response from a API Extension hosted in an AWS Lambda must contain a valid array for the `actions` property - i.e. undefined is not allowed

(This is an addition to a previous PR: https://github.com/commercetools/commercetools-adyen-integration/pull/297)

### DIFF
--- a/extension/src/lambda.js
+++ b/extension/src/lambda.js
@@ -9,10 +9,9 @@ exports.handler = async function (event) {
 
     return {
       responseType: paymentResult.success ? 'UpdateRequest' : 'FailedValidation',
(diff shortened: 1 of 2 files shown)


jamiegardiner:
@ahmetoz sorry, noticed that the previous PR still contained a small bug, so have raised this PR to fix

ahmetoz:
@jamiegardiner https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v4.0.2 the fix is released, thanks!

Issue #313, opened 2020-06-09
Support manual-capture


ahmetoz:
If a transaction of type capture is added to the payment and there is at least 1 successful authorization transaction then Adyen capture should be executed and transaction status changed. The assumption is that the user creates a transaction with the state "Initial".

https://docs.adyen.com/checkout/capture#manual-capture

PR #314, opened 2020-06-09
implement manual-capture.handler.js

ahmetoz:
#313 

#### TODOs:

- [x] Ensure the changes does not affect the notification module (before Charge/Pending transaction is not added) 
- [ ] Try to add an integration test for one of the open banking payment methods (redirect is blocking in here)
- [x] Discuss the case when there is no 'Authorisation' transaction (currently there is no check for this)


extension/test/integration/manual-capture.spec.js
```diff
@@ -0,0 +1,89 @@
...
+            originalReference: payment.transactions[0].interactionId,
+            reference: 'YOUR_UNIQUE_REFERENCE'
+          })
+        }
+      ])
+
+    expect(statusCode).to.be.equal(200)
+
+    expect(chargedPayment.transactions).to.have.lengthOf(2)
+    const transaction = chargedPayment.transactions[1]
```
lojzatran, line 85: Do you want to check interface interactions as well?

ahmetoz: You're right, It looks like I forgot it when going back and forth between unit and integration tests 😄

ahmetoz:
@lojzatran What do you think about that point in the todo list? 
> Discuss the case when there is no 'Authorisation' transaction (currently there is no check for this)

lojzatran:
> @lojzatran What do you think about that point in the todo list?
> 
> > Discuss the case when there is no 'Authorisation' transaction (currently there is no check for this)

We can check it, I don't see a big deal here.
But what is the value of the check? Theoretically the transaction should always be there as in all our handlers, we create 'Authorisation' transaction in the end. And if it's not there, then it doesn't matter, right? If the payment is not valid for capture, Adyen will return this fact to us anyway. What do you think is the benefit of having the check for 'Authorisation' transaction?

ahmetoz:
> > @lojzatran What do you think about that point in the todo list?
> > > Discuss the case when there is no 'Authorisation' transaction (currently there is no check for this)
> 
> We can check it, I don't see a big deal here.
> But what is the value of the check? Theoretically the transaction should always be there as in all our handlers, we create 'Authorisation' transaction in the end. And if it's not there, then it doesn't matter, right? If the payment is not valid for capture, Adyen will return this fact to us anyway. What do you think is the benefit of having the check for 'Authorisation' transaction?
@lojzatran 
Nothing different than what I thought, so for now I will not add this check, as it's not true in this scenario, because manual-capture is not working synchronously, it's receiving the request and asynchronously returning with notification. I asked you because maybe you could see a different point, thanks for the discussion.

lojzatran:
> > > @lojzatran What do you think about that point in the todo list?
> > > > Discuss the case when there is no 'Authorisation' transaction (currently there is no check for this)
> > 
> > 
> > We can check it, I don't see a big deal here.
> > But what is the value of the check? Theoretically the transaction should always be there as in all our handlers, we create 'Authorisation' transaction in the end. And if it's not there, then it doesn't matter, right? If the payment is not valid for capture, Adyen will return this fact to us anyway. What do you think is the benefit of having the check for 'Authorisation' transaction?
> 
> @lojzatran
> Nothing different than what I thought, so for now I will not add this check, as it's not true in this scenario, because manual-capture is not working synchronously, it's receiving the request and asynchronously returning with notification. I asked you because maybe you could see a different point, thanks for the discussion.

Let's discuss also with Roman later :) He might have also some valuable inputs.

extension/test/integration/manual-capture.spec.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/314/commits/c8accf66ff0c5f9884ae782348f56dc59e9c749d

Issue #317, opened 2020-06-10
Amount digits and currencies on tests need to be checked and fixed


ahmetoz:
Here, for example, the sample below, the amount planned is 10 EUR (1000, EUR), but makePayment request is (1000, USD), should it be validated? 🤔 

And make sure it's always using the same standards: 
https://docs.commercetools.com/http-api-types.html#money
https://docs.adyen.com/development-resources/currency-codes

https://github.com/commercetools/commercetools-adyen-integration/blob/60875df6cce45015c6aa4531b91f1db13e8af068/extension/test/integration/make-payment.handler.spec.js#L22-L42

Issue #80, opened 2019-10-08
Use lockfile-lint to prevent injection of bad packages


lojzatran:
For understanding read this article: https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/

Investigate lockfile-lint and use it in both extension and notification modules

ahmetoz:
the solution is merged into #320

PR #320, opened 2020-06-15
Cleanup the project before web-components migration (prepare v5.0.0)

ahmetoz:
### Todo List:
 
- [X] Cleanup existing files before web-components migration
- [X] Sync with current master
- [X] Update deps, remove vulnerabilities
- [X] #80 
- [ ] Unit tests depend on the configs (remove the dependency) 
- [x] Do we need python and git in Dockerfile, and why (can we optimize the docker)?
   > We're using bunyan and bunyan needs python to compile and run `dtrace`, so it's needed.
- [x] #317 
- [x] Do coverage analysis, add min coverage to build
- [x] Force eslint for every commit (run on build)
- [ ] Refactor for to improve quality, define uncovered parts of the project


extension/package.json
```diff
@@ -45,39 +39,35 @@
...
+    "mocha": "^8.0.1",
     "ngrok": "^3.2.7",
+    "nyc": "^15.1.0",
     "nock": "^12.0.3",
-    "nyc": "^15.0.0",
     "proxyquire": "^2.1.3",
-    "sinon": "^9.0.1",
-    "start-server-and-test": "1.10.11"
+    "sinon": "^9.0.2",
+    "start-server-and-test": "1.11.0"
```
lojzatran, line 57: I think you can remove this dep. I don't see it's being used anywhere and before it was because of cypress with command `server-test`.

Issue #298, opened 2020-06-02
Add e2e UI tests


lojzatran:
Add e2e tests for following processes:
- credit card without 3DS
- credit card with 3DS
- Klarna

PR #321, opened 2020-06-16
298 e2e tests

lojzatran:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/298


extension/test/e2e/credit-card.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    { name: 'Visa Electron', creditCardNumber: '4001 0200 0000 0009' },
+    { name: 'V Pay', creditCardNumber: '4013 2500 0000 0000 006' },
+  ]
+
+  // note: ngrok should be restarted for every test case, otherwise there will be
+  // 429 Too Many Requests error. This is due to the limit of maximum opened HTTP connections,
+  // which is 40 connections at the same time.
+  beforeEach(async () => {
+    const fileServer = new nodeStatic.Server()
+    routes['/make-payment-form'] = async (request, response) => {
```
ahmetoz, line 43: why the file name is not credit-card-payment-form? (make-payment could be any payment right ?)

extension/test/e2e/credit-card.spec.js
```diff
@@ -0,0 +1,118 @@
...
+            }
+          }
+        }
+
+        const { body: payment } = await ctpClient.create(ctpClient.builder.payments, paymentDraft)
+        const { getOriginKeysResponse: getOriginKeysResponseString } = payment.custom.fields
+        const getOriginKeysResponse = await JSON.parse(getOriginKeysResponseString)
+
+        // Make payment:
+        const makePaymentFormPage = new MakePaymentFormPage(page, baseUrl)
```
ahmetoz, line 99: Why this is not called `CardComponent` as Adyen calls it that way? https://docs.adyen.com/payment-methods/cards/web-component

Think those e2e tests as documentation. It is always good to use the right naming and easy to reason code.

do we need to pass a page, browser, static file router ext in here, maybe it should be part of the component, otherwise, there is no value doing a decoupling as you did. So CardComponent will manage everything related to client-side and fixture will only do the extension call part (like the customer will do the same )?  

new CardComponent() ? 

Here I guess we need to define a strategy, another option for me we could also remove decoupling in here and write the puppeteer operations on the same test instead of moving them into the page objects folder which is confusing. If we are doing decoupling I would ask for a better implementation, so this means you need to structure the tests with this goal in mind.

extension/package.json
```diff
@@ -61,8 +62,10 @@
     "mocha": "^7.1.0",
     "ngrok": "^3.2.7",
     "nock": "^12.0.3",
+    "node-static": "^0.7.11",
     "nyc": "^15.0.0",
     "proxyquire": "^2.1.3",
+    "puppeteer": "^3.2.0",
```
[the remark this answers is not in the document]
lojzatran: Wow they're releasing new versions like Google Chrome 😄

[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/43693cf17084053367cce5fcf1d627113b4fba94#diff-0491f1a79162851367d16a7b21c548a7R68

extension/test/e2e/fixtures/make-payment-form.html
```diff
@@ -0,0 +1,289 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <title>Make Payment</title>
+        <link rel="stylesheet"
+              href="https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.6.1/adyen.css"
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/43693cf17084053367cce5fcf1d627113b4fba94

extension/test/e2e/credit-card-redirect.spec.js
```diff
@@ -0,0 +1,174 @@
...
+const { routes } = require('../../src/routes')
+const c = require('../../src/config/constants')
+const httpUtils = require('../../src/utils')
+const { assertPayment } = require('./e2e-test-utils')
+const MakePaymentFormPage = require('./pageObjects/CreditCardMakePaymentFormPage')
+const RedirectPaymentFormPage = require('./pageObjects/RedirectPaymentFormPage')
+const CreditCardRedirectPage = require('./pageObjects/CreditCardRedirectPage')
+
+// Flow description: https://docs.adyen.com/checkout/3d-secure/redirect-3ds2-3ds1/web-component
+describe('credit-card-payment-redirect', () => {
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/f30706462dff19110353d22e866f7accfa4c64ae

extension/test/e2e/pageObjects/CreditCard3dsV2Page.js
```diff
@@ -0,0 +1,19 @@
+const { executeInAdyenIframe } = require('../e2e-test-utils')
+
+module.exports = class CreditCard3dsV2Page {
+
+  constructor (page, baseUrl) {
+    this.page = page
+    this.baseUrl = baseUrl
+  }
+
+  async finish3DSV2Payment () {
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/f30706462dff19110353d22e866f7accfa4c64ae#diff-4c3189dc9ec0bbb1367f7a491d438824R9

extension/test/e2e/pageObjects/CreditCardMakePaymentFormPage.js
```diff
@@ -0,0 +1,19 @@
+const { executeInAdyenIframe } = require('../e2e-test-utils')
+const MakePaymentFormPage = require('./MakePaymentFormPage')
+
+module.exports = class CreditCardMakePaymentFormPage extends MakePaymentFormPage {
+  async goToThisPage () {
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/ca36b206c94043b9acc47ec8691809b28e127a40

extension/test/e2e/e2e-test-utils.js
```diff
@@ -0,0 +1,46 @@
...
+  for (const frame of page.mainFrame().childFrames()) {
+    const elementHandle = await frame.$(selector)
+    if (elementHandle) {
+      await executeFn(elementHandle)
+      break
+    }
+  }
+}
+
+function assertPayment (payment, finalInterfaceInteractionFromAdyen = 'submitAdditionalPaymentDetails') {
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/c676a687290a8c971b94154001e1429e48740541

extension/test/e2e/credit-card.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  })
+
+  creditCards.forEach(({
+                         name, creditCardNumber, creditCardDate = '03/30', creditCardCvc = '737'
+                       }) => {
+    // eslint-disable-next-line no-template-curly-in-string
+    it(`when issuer is ${name} and number is ${creditCardNumber}, ` +
+      'it should successfully finish the payment with credit card',
+      async () => {
+        const page = await browser.newPage()
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/ca36b206c94043b9acc47ec8691809b28e127a40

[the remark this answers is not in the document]
lojzatran: `also here it looks like all those duplicate codes could be collected in a parent credit-card file so it would be easy to focus the important parts of the test, wdyt ?`

This file doesn't have any parent class and it's quite short now. So I guess it's ok?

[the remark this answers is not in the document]
ahmetoz: Yes, it seems you got the idea, parent class, or a common unit class so we could change it easily.

ahmetoz:
@lojzatran  
> It's a huge increase in time for sure, but we're testing lots of payment methods and different credit cards.. 

Actually do we need to test all banks? I don't think so. What is the idea here?

lojzatran:
> @lojzatran
> 
> > It's a huge increase in time for sure, but we're testing lots of payment methods and different credit cards..
> 
> Actually do we need to test all banks? I don't think so. What is the idea here?

The idea is to make sure that the banks work the same. Actually when testing it, I discovered that they were not working without explicitly asking Adyen to activate :) So we know some new things here and now we can tell our customers that we have these banks tested and working. But the flow is the same for all of them in the end.

extension/test/e2e/credit-card.spec.js
ahmetoz: so reusability for puppeteer operations is not a problem in here, it should be clean. check the design below, the code is passing page as a constructor but not using anywhere else, which is actually a state of the MakePaymentFormPage which is managed by there, maybe same for baseUrl. 

https://github.com/commercetools/commercetools-adyen-integration/blob/ca36b206c94043b9acc47ec8691809b28e127a40/extension/test/e2e/credit-card.spec.js#L70-L73

So check this comment again: 
> do we need to pass a page, browser, static file router ext in here, maybe it should be part of the component, otherwise, there is no value doing a decoupling as you did.

Here another case, so this should be a method. 
https://github.com/commercetools/commercetools-adyen-integration/blob/ca36b206c94043b9acc47ec8691809b28e127a40/extension/test/e2e/credit-card-3ds-native.spec.js#L133-L136

So my comment is neither implemented nor understood.

> If we are doing decoupling I would ask for a better implementation, so this means you need to structure the tests with this goal in mind.

Also, please better names... you are not using the domain language in the naming, in the end tests are not only for something to evaluate the code, IMO especially e2e tests should also be meaningful. 

Let me know how we could improve the existing flow especially readability in tests. (should not be hard to do for you)

ahmetoz:
> > @lojzatran
> > > It's a huge increase in time for sure, but we're testing lots of payment methods and different credit cards..
> > 
> > 
> > Actually do we need to test all banks? I don't think so. What is the idea here?
> 
> The idea is to make sure that the banks work the same. Actually when testing it, I discovered that they were not working without explicitly asking Adyen to activate :) So we know some new things here and now we can tell our customers that we have these banks tested and working. But the flow is the same for all of them in the end.


Solving methods activation problem in Adyen merchant account with our tests should not be the goal here, you faced this issue and you asked to activate those, in the end maybe if we test for 1 credit card instead of 12 will reduce the build time from 8 minutes to 2 minutes, which looks like an important thing for me for development experience. Here, I guess customers are not interested specifically in what we are testing every credit card type in e2e, they will just read our documentation and understand we are supporting all credit card payments, same for other payment methods also when you consider other payment types, we don't have e2e for every payment type, right?

lojzatran:
> > > @lojzatran
> > > > It's a huge increase in time for sure, but we're testing lots of payment methods and different credit cards..
> > > 
> > > 
> > > Actually do we need to test all banks? I don't think so. What is the idea here?
> > 
> > 
> > The idea is to make sure that the banks work the same. Actually when testing it, I discovered that they were not working without explicitly asking Adyen to activate :) So we know some new things here and now we can tell our customers that we have these banks tested and working. But the flow is the same for all of them in the end.
> 
> Solving methods activation problem in Adyen merchant account with our tests should not be the goal here, you faced this issue and you asked to activate those, in the end maybe if we test for 1 credit card instead of 12 will reduce the build time from 8 minutes to 2 minutes, which looks like an important thing for me for development experience. Here, I guess customers are not interested specifically in what we are testing every credit card type in e2e, they will just read our documentation and understand we are supporting all credit card payments, same for other payment methods also when you consider other payment types, we don't have e2e for every payment type, right?

Our goal is to check if the credit card payment is working. The more tests we have, the better it is to argue later if sth doesn't work that it is not our fault. And now I remember that when testing credit cards, one card from that group was not working as expected as it has a bit different flow. I reported this to Adyen and they admit documentation is a bit misleading.
I would test also other methods if we have more time, but here we have a promise from Adyen that they will work out of the box.
But more tests is of course more problems - slower times and sometimes tests are timing out because response is slow. So if you don't see the benefits, I can remove them.

ahmetoz:
> one card from that group was not working as expected as it has a bit different flow. I reported this to Adyen and they admit documentation is a bit misleading.

@lojzatran I guess you could keep this but the others actually are the same, so those are not important to me, just a different credit card issue and even a problem occur it will be Adyen related not extension module because we are following their contracts so we could not do much even it does not work, it will just block our build, right? I am interested to know the real value, not some exceptional cases that we can not fix, when you keep test times are slow, probably we will get bored to wait to get results, in pay one example I am bored to wait for tests, also it's different because pay one has lots of flows so it's not doing the same thing again and again. What do you think now about to drop those tests and only keep the critical ones? (same for 3Ds methods, maybe one bank might be enough).

extension/test/e2e/credit-card.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    { name: 'Hipercard', creditCardNumber: '6062 8288 8866 6688' },
+    { name: 'JCB', creditCardNumber: '3569 9900 1009 5841' },
+    { name: 'Mastercard', creditCardNumber: '5101 1800 0000 0007' },
+    { name: 'UATP', creditCardNumber: '1354 1001 4004 955', creditCardDate: '06/22' },
+    { name: 'VISA', creditCardNumber: '4166 6766 6766 6746' },
+    { name: 'Visa Electron', creditCardNumber: '4001 0200 0000 0009' },
+    { name: 'V Pay', creditCardNumber: '4013 2500 0000 0000 006' },
+  ]
+
+  // note: ngrok should be restarted for every test case, otherwise there will be
```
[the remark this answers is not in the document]
ahmetoz: here it could be nice to remove duplicate comments and explain this in one place, please also add in to comment your description of why it has a limitation which is the main reason why you have this workaround. 

> Limit applies because we have a free program: https://ngrok.com/pricing

lojzatran:
> > one card from that group was not working as expected as it has a bit different flow. I reported this to Adyen and they admit documentation is a bit misleading.
> 
> @lojzatran I guess you could keep this but the others actually are the same, so those are not important to me, just a different credit card provider/bank and even a problem occur it will be Adyen related not extension module because we are following their contracts so we could not do much even it does not work, it will just block our build, right? I am interested to know the real value, not some exceptional cases that we can not fix when you keep test times slow without a doubt we will get bored to wait to get results, in pay one example I am bored to wait for tests to finish in 15 minute, also it's different because pay one has lots of flows so it's not doing the same thing again and again. What do you think now about to drop those tests and only keep the critical ones? (same for 3Ds methods, maybe one bank might be enough).

Yeah let's do it. It also makes developer's life easier :D https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/385546b67eb2a64afe8a39ffdaed38fc1321b6f0

extension/test/e2e/credit-card.spec.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/323 Thanks :) Now I understand the comment. I should split the big test method into smaller methods.

lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/2a23d0c0c3665f135a628a3311dd04129ca838e6

extension/test/e2e/credit-card-3ds-native.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    await iTSetUp.stopRunningServers()
+    await browser.close()
+  })
+
+  creditCards.forEach(({
+                         name, creditCardNumber, creditCardDate = '03/30', creditCardCvc = '737'
+                       }) => {
+    // eslint-disable-next-line no-template-curly-in-string
+    it(`when credit card issuer is ${name} and credit card number is ${creditCardNumber}, ` +
+      'it should successfully finish the payment with 3DS native authentication flow',
```
ahmetoz, line 63:
```suggestion
      'then it should successfully finish the payment with 3DS native authentication flow',
```

extension/test/e2e/credit-card-3ds-native.spec.js
```diff
@@ -0,0 +1,178 @@
...
+      async () => {
+        const baseUrl = process.env.API_EXTENSION_BASE_URL
```
ahmetoz, line 65: Refactoring point: All those baseUrls should be managed by config.load.

extension/test/e2e/credit-card-3ds-native.spec.js
```diff
@@ -0,0 +1,178 @@
...
+          creditCardDate,
+          creditCardCvc,
+          payment
+        })
+
+        const paymentAfterIdentifyShopper = await identifyShopper({
+          payment: paymentAfterMakePayment, browserTab, baseUrl, getOriginKeysResponse
+        })
+
+        const paymentAfterAuthentication = await performAuthenticationFlow({
```
ahmetoz, line 86:
```suggestion
        const paymentAfterAuthentication = await challengeShopper({
```

extension/test/e2e/credit-card-3ds-redirect.spec.js
```diff
@@ -0,0 +1,146 @@
+const nodeStatic = require('node-static')
```
ahmetoz, line 1: why not simply using my implementation instead of changing it, I spend time and explained better the tricky redirect part 😉

extension/test/e2e/credit-card-3ds-native.spec.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/3357ae9869afff108125e98633761970346470d9

extension/test/e2e/credit-card-3ds-native.spec.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/3357ae9869afff108125e98633761970346470d9

extension/test/e2e/credit-card-3ds-native.spec.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/321/commits/95fd82534acac4aa5cc254219b22a561403e9889

PR #323, opened 2020-06-17
Try to refactor `credit-card-3ds-redirect.spec.js`

ahmetoz:
Let’s decide what could be next for other files. I guess I made my point clear.

4-5 lines code, nothing complex, using the same terms in the documentation, explains the tricky return-URL part with an example representation data. Otherwise, it's not easy to understand what's happening without debugging the existing code.

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "check-coverage": "nyc check-coverage --statements 97",
-    "test": "npm run lint && nyc mocha --exit --timeout 30000 --full-trace test/**/*.spec.js && npm run check-coverage",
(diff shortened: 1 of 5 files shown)


extension/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "check-coverage": "nyc check-coverage --statements 97",
-    "test": "npm run lint && nyc mocha --exit --timeout 30000 --full-trace test/**/*.spec.js && npm run check-coverage",
+    "test": "npm run lint && npm run unit && npm run integration && npm run e2e && npm run check-coverage",
```
lojzatran, line 8: I think this will get the coverage only from `npm run e2e`, which is not 97%.

PR #327, opened 2020-06-18
add getPaymentMethods and getOriginKeys into new integration guide.

ahmetoz:
@lojzatran  Please check the changes and provide input asap before doing other flows, so we could find an easy to follow documentation for customers.

Note: The new integration guide is not complete and could be changed.


lojzatran:
Might be good to assign a review to someone who never worked with this before to see if they understand the process. WDYT?

ahmetoz:
> Might be good to assign a review to someone who never worked with this before to see if they understand the process. WDYT?

Yes but we don't have time for this, we could ask it later in the latest version, so here it would be nice to discuss before doing other works, it will reduce additional code review problems later.

lojzatran:
The structure is good for me. I would put the steps in one page. It's easier to read as you don't have to click the links back and forth if you want to check other steps (like Adyen has with their docs).

extension/docs/Step1-getPaymentMethods.md
```diff
@@ -0,0 +1,109 @@
...
+    "details":[...],
+    "name":"SEPA Direct Debit",
+    "type":"sepadirectdebit"
+  },
+  ...
+  ]
+}
+```
+
+CTP payment representation:
```
lojzatran, line 69: I think we can remove unnecessary data from the representation which is not focus of this step. For example: amountPlanned or interfaceInteractions.

Issue #355, opened 2020-07-06
Update documentation of Notification module


ahmetoz:
After changing all documents in the extension module with PR: #329 
It's necessary to update the notification module's documentation to make it consistent with style and content.

PR #357, opened 2020-07-07
Update notification docs

ahmetoz:
#355


notification/docs/DeploymentGuide.md
```diff
@@ -0,0 +1,69 @@
...
+  - [Other Configurations](#other-configurations)
+- [Deployment](#deployment)
+  - [Docker](#docker)
+    - [Running the Docker image](#running-the-docker-image)
+  - [AWS Lambda](#aws-lambda)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Environment variables
+Following environment variables must be provided in order to run the notification module.
```
butenkor, line 19:
```suggestion
Following environment variables are supported.
```

notification/docs/DevelopmentGuide.md
```diff
@@ -1,45 +1,27 @@
...
-
-## Run Tests
-
-- Execute `npm run unit-test` to run [Unit tests](../test/unit)
-- Execute `npm run integration-test` to run [Integration tests](../test/integration) (all required environment variables must be set)
-
-Execute `npm run test` to run all tests
-
-In the end, a code coverage report will be printed.
+> You need to set [environment variables](./DeploymentGuide.md#environment-variables) to be able to run integration tests.
```
butenkor, line 27: Does it work with any empty CT project or you need a special / preconfigured one?

ahmetoz: Yes, tests will bootstrap the required resources. 

But I have checked again it' like we did not adopt the IT of notification with the new payment types, I will try to update those too 🤦 https://github.com/commercetools/commercetools-adyen-integration/blob/355-notification-docs/notification/test/resources/payment-custom-type.json

notification/docs/DeploymentGuide.md
ahmetoz: I will keep as is, I could not get why we need to say environment variables `are supported` 🤔

notification/docs/DevelopmentGuide.md
ahmetoz: The issue is created, I will handle this as a separate PR: https://github.com/commercetools/commercetools-adyen-integration/issues/365

notification/docs/DeploymentGuide.md
butenkor: Because some variables are mandatory and some not and you state upfront that all of them are mandatory (`must`).

PR #370, opened 2020-07-21
Fix serialize-error package import

ethann:
Since version 5.0.0 serialize-error provides 2 methods:
- serializeError
- deserializeError


butenkor:
Hi @ethann Thanks for your contribution but why do you need to import deserializeError if it is not used?

ethann:
That's correct! Thanks for pointing it out. Fixed.

PR #406, opened 2020-08-19
add note for http extension timeout limit.

ahmetoz:


extension/docs/DeploymentGuide.md
```diff
@@ -61,6 +61,7 @@ If you don't have the commercetools OAuth credentials,[create a commercetools AP
 Resources below are required for the extension module to operate correctly. Resources that ***will be automatically created*** by the extension module in your commercetools project.
 
 1. [The commercetools HTTP API Extension pointing to Adyen extension module](../resources/api-extension.json)
+> It's required that the HTTP API Extension timeout limit be increased to 10000 milliseconds (default is 2000). Please contact Support via the commercetools [support portal](https://support.commercetools.com/) and provide the region, project key, and use case to increase timeout to 10000.
```
lojzatran, line 64:
```suggestion
> It's required that the HTTP API Extension timeout limit is increased to 10000 milliseconds (default is 2000). Please contact Support via the commercetools [support portal](https://support.commercetools.com/) and provide the region, project key, and use case to increase timeout to 10000.
```

lojzatran (approved):

extension/docs/DeploymentGuide.md
butenkor: I would also add that after limit increase an API extension timeout needs to be updated over API too: https://docs.commercetools.com/http-api-projects-api-extensions#set-timeoutinms

ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/406/commits/f8e5eee950687475b7591bda9958949670c20ba1

PR #450, opened 2020-10-19
Fix test for klarna

lojzatran:
Klarna test was failing because some HTML elements are not present anymore on the page.


ahmetoz (approved):
They are changing too often :trollface:

Issue #461, opened 2020-10-28
Prevent on payment creation inconsistent amounts provided on makePaymentRequest and amountPlaned


lojzatran:
The problem: it is possible to create a payment with different values in `amountPlanned` and `makePaymentRequest custom field`. The problem is we're checking interface interaction for `makePaymentRequest`. If there's no interface interaction yet, we allow the value to be not matching. This leads to inconsistent payments that always fails.

Solution: `makePaymentRequest custom field` should be checked with `amountPlanned` if there's no interface interactions.

PR #462, opened 2020-10-28
#461 Prevent on payment creation inconsistent amounts provided on makePaymentRequest and amountPlaned

lojzatran:
The problem: it is possible to create a payment with different values in `amountPlanned` and `makePaymentRequest custom field`. The problem is we're checking interface interaction for `makePaymentRequest`. If there's no interface interaction yet, we allow the value to be not matching. This leads to inconsistent payments that always fails.

Solution: `makePaymentRequest custom field` should be checked with `amountPlanned` if there's no interface interactions.

Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/461

### DIFF
--- a/extension/src/validator/error-messages.js
+++ b/extension/src/validator/error-messages.js
@@ -4,6 +4,7 @@ module.exports = {
   MAKE_PAYMENT_REQUEST_INVALID_JSON: 'makePaymentRequest does not contain valid JSON.',
   // eslint-disable-next-line max-len
   SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON: 'submitAdditionalPaymentDetailsRequest does not contain valid JSON.',
-  AMOUNT_PLANNED_NOT_SAME: 'amountPlanned field must be the same as the amount in makePayment request',
+  AMOUNT_PLANNED_NOT_SAME: 'amountPlanned field must be the same as the amount in '
+    + 'makePayment request in the interface interactions or in the custom field',
   MAKE_PAYMENT_REQUEST_MISSING_REFERENCE: 'Required "reference" field is missing in makePaymentRequest.'
 }
--- a/extension/src/validator/error-messages.js
+++ b/extension/src/validator/error-messages.js
@@ -5,6 +5,6 @@ module.exports = {
   // eslint-disable-next-line max-len
   SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON: 'submitAdditionalPaymentDetailsRequest does not contain valid JSON.',
   AMOUNT_PLANNED_NOT_SAME: 'amountPlanned field must be the same as the amount in '
-    + 'makePayment request in the interface interactions or in the custom field',
+    + 'makePaymentRequest in the interface interactions or makePaymentRequest  in the custom field',
   MAKE_PAYMENT_REQUEST_MISSING_REFERENCE: 'Required "reference" field is missing in makePaymentRequest.'
 }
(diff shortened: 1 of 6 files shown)


extension/src/validator/error-messages.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   MAKE_PAYMENT_REQUEST_INVALID_JSON: 'makePaymentRequest does not contain valid JSON.',
   // eslint-disable-next-line max-len
   SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON: 'submitAdditionalPaymentDetailsRequest does not contain valid JSON.',
-  AMOUNT_PLANNED_CHANGE_NOT_ALLOWED: 'amountPlanned field cannot be changed after makePayment request is made to Adyen',
+  AMOUNT_PLANNED_NOT_SAME: 'amountPlanned field must be the same as the amount in '
+    + 'makePayment request in the interface interactions or in the custom field',
```
ahmetoz, line 8:
```suggestion
    + 'makePayment request in the interface interactions or makePaymentRequest  in the custom field',
```

extension/src/validator/error-messages.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   MAKE_PAYMENT_REQUEST_INVALID_JSON: 'makePaymentRequest does not contain valid JSON.',
   // eslint-disable-next-line max-len
   SUBMIT_ADDITIONAL_PAYMENT_DETAILS_REQUEST_INVALID_JSON: 'submitAdditionalPaymentDetailsRequest does not contain valid JSON.',
-  AMOUNT_PLANNED_CHANGE_NOT_ALLOWED: 'amountPlanned field cannot be changed after makePayment request is made to Adyen',
```
ahmetoz, line 7: so it can be changed ?

extension/test/unit/payment-handler.spec.js
```diff
@@ -170,42 +168,37 @@ describe('payment-handler::execute', () => {
       expect(response.data.actions).to.deep.equal([])
     })
 
-    it('when amountPlanned is updated and multiple makePayment exists, ' +
```
ahmetoz, line 173: should not be those tests valid after your change, why updating it?

extension/src/validator/validator-builder.js
```diff
@@ -32,18 +32,27 @@ function withPayment (paymentObject) {
...
-        if (oldAmount !== newAmount || oldCurrencyCode !== newCurrencyCode)
-          errors.amountPlanned = errorMessages.AMOUNT_PLANNED_CHANGE_NOT_ALLOWED
+      if (makePaymentRequestInterfaceInteraction)
+        amount = JSON.parse(makePaymentRequestInterfaceInteraction.fields.request).amount
+      else {
+        const makePaymentRequestString = paymentObject.custom
+          && paymentObject.custom.fields
+          && paymentObject.custom.fields.makePaymentRequest
+        if (makePaymentRequestString)
+          amount = JSON.parse(makePaymentRequestString).amount
```
ahmetoz, line 46: what happens when amount is not set in makePaymentRequest ? 🤔

ahmetoz (commented):
In general, it look good, I have some comments.

extension/src/validator/validator-builder.js
lojzatran: It's a required parameter. If it's not set, Adyen request will fail. https://docs.adyen.com/checkout/components-web#step-3-make-a-payment

extension/test/unit/payment-handler.spec.js
lojzatran: They are not valid anymore as now we additionally check for both `makePaymentRequest` interface interaction and `makePaymentRequest` custom field.

ahmetoz (approved):

Issue #466, opened 2020-10-30
Identify the application to Adyen


lojzatran:
According to the recommendations of Adyen (https://docs.adyen.com/development-resources/building-adyen-solutions) an integration app should include headers to identify itself.

PR #467, opened 2020-10-30
#466 add application info to the requests

lojzatran:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/466


ahmetoz (approved):
Looks good, testing would be better without duplicated code, but this is ok for now.

Issue #399, opened 2020-08-17
Refund part of captured amount


christianandersson:
Perhaps this will not be an issue when the next major is released.. However, we've been using a build of the Web-components branch for some time. How would you go about refunding a partial amount? The cancellation action calls the cancelOrRefund endpoint at Adyen here:
 
https://github.com/commercetools/commercetools-adyen-integration/blob/support-adyen-web-components/extension/src/service/web-component-service.js#L24

This API call will cancel everything that has been captured and not just the amount set on the transaction. See docs:

https://docs.adyen.com/checkout/cancel-or-refund#cancel-or-refund-a-payment

I think you'd have to call the refund endpoint at Adyen to refund part of the payment.

https://docs.adyen.com/checkout/refund

butenkor:
Hi @christianandersson, unfortunately In our first MVP we did not consider supporting partial charges or refunds. We might add it though soon (regarding timing please create support ticket here http://support.commercetools.com). If you are in a rush you can also consider providing a PR, where depending on the refund transaction amount the extension would do a partial refund instead of cancelOrRefund.

christianandersson:
Thanks you @butenkor 
There 's probably no way for a retailer to get around such a limitation, unless they can handle refunds in the Adyen payment portal. We might patch the current version and add support for this, if so, we can submit a PR.

PR #485, opened 2020-11-24
#399 notification module - get matching transactions by interactionId

lojzatran:
todo:
- [x] add test for multiple refund notifications
- [ ] manually test the whole process of refund/cancellation to have a success notification
- [x] update README

### DIFF
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -64,8 +64,8 @@ async function updatePaymentWithRepeater (payment, notification, ctpClient) {
           + ` max retries. Error: ${JSON.stringify(serializeError(err))}`)
       }
       /* eslint-disable-next-line no-await-in-loop */
-      currentPayment = await ctpClient.fetchById(ctpClient.builder.payments, currentPayment.id)
-      currentPayment = currentPayment.body.results[0] // eslint-disable-line prefer-destructuring
+      const response = await ctpClient.fetchById(ctpClient.builder.payments, currentPayment.id)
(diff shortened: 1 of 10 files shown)


notification/src/handler/notification/notification.handler.js
```diff
@@ -64,8 +64,8 @@ async function updatePaymentWithRepeater (payment, notification, ctpClient) {
           + ` max retries. Error: ${JSON.stringify(serializeError(err))}`)
       }
       /* eslint-disable-next-line no-await-in-loop */
-      currentPayment = await ctpClient.fetchById(ctpClient.builder.payments, currentPayment.id)
-      currentPayment = currentPayment.body.results[0] // eslint-disable-line prefer-destructuring
+      const response = await ctpClient.fetchById(ctpClient.builder.payments, currentPayment.id)
+      currentPayment = response.body // eslint-disable-line prefer-destructuring
       currentVersion = currentPayment.version
```
JudeNiroshan, line 69: is it guaranteed that we will never have an undefined error for `currentPayment.version`?

JudeNiroshan (approved):

notification/src/handler/notification/notification.handler.js
lojzatran: Nothing is guaranteed 100% in life, and it's the same for this case. It could be that the payment does not exist anymore for some reason and we would get null for `currentPayment`. However this would mean that the payment was deleted in between (this code is in the place where you have 409 error) and I cannot think of a reason why it could happen. It's just theory. So for now, I'd say it's ok to have it like this. WDYT?

Issue #50, opened 2019-04-16
Add liveness and readiness probes


LEQADA:

ahmetoz:
@LEQADA  could you please clarify that? Did you mean health endpoints for the modules?

LEQADA:
We created k8s deployment yamls for both modules here https://github.com/commercetools/commercetools-adyen-integration/tree/k8s-deployment

The deployment doesn't have readiness and liveness probes. It needs to be discussed how they'll be implemented.

PR #495, opened 2020-12-16
Add health endpoint to notification module

ahmetoz:
Fixes: #50


JudeNiroshan:
can you also add the `/health` endpoint for the k8s readiness and liveness annotations in the YAML files?

ahmetoz:
> @ahmetoz can you also add the `/health` endpoint for the k8s readiness and liveness annotations in the YAML files?

> @JudeNiroshan you're right, could you check this commit: https://github.com/commercetools/commercetools-adyen-integration/pull/495/commits/7f69e75049658230d044a162770d844797863a85

deployment-examples/gcp/extension/k8s/values.yaml
```diff
@@ -17,12 +17,3 @@ ingress:
   annotations:
     kubernetes.io/ingress.class: 'gce'
     kubernetes.io/ingress.allow-http: 'true'
-
```
JudeNiroshan, line 20: Why it is removed? Shouldn't we use our health endpoint? Otherwise, there is no need to have health endpoint from the application.

ahmetoz: It's the default probe url in the helm chart, I also added the link, you might check.

JudeNiroshan (approved):

PR #496, opened 2020-12-16
Adapt to distroless images and use multi stage build

ahmetoz:
* Details: https://github.com/GoogleContainerTools/distroless

Notes: 
- Also, the docker image size is reduced from 167 MB to 107 MB.
- Benefits: https://www.thoughtworks.com/radar/techniques/distroless-docker-images


ahmetoz:
Decided to not use.

PR #508, opened 2021-01-05
Add prettier.js 

### LINKED ISSUE #507: Using prettier code formatter
**Situation**

There are different code formatting styles in JS and different contributors could have different formatting styles.

**Complication**

Code doesn't look consistent when there's not one style.

**Solution**

Use the Prettier library to enforce one code formatting style.

**Additional info**

Some might ask why using prettier when we already have eslint? The reason is eslint does not provide many formatting rules. Therefore inconsistencies in style could still happen. https://prettier.io/docs/en/comparison.html

It is recommended to use eslint in combination with prettier to achieve the best results. Just be aware that some rules might conflict with each other. Maybe check this as the solution: https://github.com/prettier/prettier-eslint

ahmetoz:
Fixes #507 

## Notes for the reviewer:

- Git history will be overridden and I could not find a proper way to avoid it. 

- Only 2 options are changed in favor of our coding styles, Details: https://prettier.io/docs/en/options.html
```
{
  "semi": false,
  "singleQuote": true
}
```

- Integrated with eslint with using the eslint config. Details:  https://prettier.io/docs/en/integrating-with-linters.html
- Integrated to the git commit hook, so formatting will be done on the git's staged changes, Details: https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged


lojzatran (approved):
LGTM 🎉

JudeNiroshan (approved):

PR #515, opened 2021-01-07
[WIP] Share codebase between extension and notification module

lojzatran:
In this PR, there are initial commits and set up to share the common codes between notification and extension modules.

I used [lerna](https://github.com/lerna/lerna) library to enable the code sharing. I did not use the default structure of lerna by putting all codes inside `./packages`, but decided to leave the structure as it is now. If we decided to go differently in the future, this could be easily changed in `lerna.json`.

One has to run `npm i` from the root package now, not from the subfolders anymore. All the common dependencies for both modules are now located in the `package.json` in the root folder. Because of this fact, I added `import/no-extraneous-dependencies` to eslint config to eliminate the possible eslint error.

I created a new folder called `commons` and put one file that is shared in both notification and extension modules. Referencing to this module is working and this module could later be used as a base for further improvements.

**TODO:**
1. Find a way to run `npm ci` with lerna.
1. Put all common codes into `commons` folder.
1. Fix tests - it seems like the shared code between extension and notification modules are not the same and thus some tests are failing.
1. Test the deployment process - currently we're deploying 2 independent docker images and it should remain this way. Also check how the versioning will work in this case.


ahmetoz:
@lojzatran is this PR still valid ?

lojzatran:
> @lojzatran is this PR still valid ?

If we ever want to use lerna, then it is valid. However, this PR is very old now and contains too many conflicts and hard to resolve. I will close it.

PR #520, opened 2021-01-13
Migrate CI/CD pipelines to github actions from travis ci

ahmetoz:
Check the action sections for more details: https://github.com/commercetools/commercetools-adyen-integration/actions

CI workflow contains the build (triggered every commit)
CD workflow contains the docker push workflow (triggered with a new tag)

Currently, the setup is exactly the same as the Travis setup.
 
I tested the docker push and it's also working as expected.


.github/workflows/cd.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: actions/checkout@v2
+      - name: Set output
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
```
salander85, line 25: Don't you need to logout when deployment was done?

ahmetoz: Check: https://github.com/commercetools/commercetools-adyen-integration/runs/1696318350?check_suite_focus=true 
Check the step:
Screenshot 2021-01-20 at 13 50 12

salander85: nice^^

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+      - run: npm run test
+        env:
+          ADYEN_SECRET_HMAC_KEY: ${{ secrets.ADYEN_SECRET_HMAC_KEY }}
+          CTP_PROJECT_KEY: adyen-integration-test
+          CTP_CLIENT_ID: ${{ secrets.CTP_CLIENT_ID }}
+          CTP_CLIENT_SECRET: ${{ secrets.CTP_CLIENT_SECRET }}
+
+  extension_build:
+    name: Extension Module
+    needs: notification_build
```
JudeNiroshan, line 36: I guess we do not need to depend on notification build here. Isn't it?

JudeNiroshan (changes requested):

.github/workflows/ci.yml
ahmetoz: In general yes, here the reason: we are using one commercetools project in both module's tests, so to avoid data conflicts I make them sequential. 
When we introduce a new testing setup we might remove the needs and they could run parallel.

JudeNiroshan: ahh ok

JudeNiroshan (approved):

salander85:
👍 approved

salander85 (approved):

PR #526, opened 2021-01-20
Update config for multitenancy

lojzatran:
**Situation:**
Currently credentials of the Adyen extension provided with environment variables (both Adyen and commercetools project credentials) in order to run the extension module.

**Complication:**
To handle multiple commercetools/Adyen projects, the module will need a more dynamic and structured configuration object (like JSON) rather than a simple string environment variable value to get multi commercetools project and Adyen merchant configurations.

**Resolution:**
Extend current single tenant setup for extension module to be able to support the multitenancy. (see: config.js)
Provide utility which will return config based on the passed adyenMerchantAccount and commercetoolsProjectKey values.
The config should be validated (fail fast if the passed config is not valid - for example, not valid JSON)
Adapt existing code base + unit tests according to the new config.

**Hints for reviewers**
1. I updated many tests files so that they use the new config structure. You don't have to review the test files.
2. Main change is extension/src/config/config.js (see its test as well - this is the only test you should review) - here the new config structure is being used - we now supply all config as JSON.
3. Another bigger change is extension/src/config/init/ensure-resources.js - we need to ensure resources for all provided ctp projects now.
4. I also fixed the error for klarna e2e test here, so after this PR is merged to master, klarna test will also be fixed. (EDIT: this change has been already merged to master)

Missing stuff:
1. Additional integration tests for multitenancy cases - will be added later in a different PR
2. Documentation updates - will be done in another PR


extension/src/ctp.js
```diff
@@ -61,11 +61,12 @@ function createCtpClient({
   })
 }
 
-function setUpClient() {
-  const ctpClient = createCtpClient(config.ctp)
+function setUpClient(ctpProjectKey) {
+  const ctpConfig = config.getCtpConfig(ctpProjectKey)
```
ahmetoz, line 65: We need to consider a cache for managing multiple client objects.

extension/src/config/config.js
```diff
@@ -1,58 +1,87 @@
-const _ = require('lodash')
+let config
 
 function getEnvConfig() {
```
ahmetoz, line 3:
```suggestion
function getModuleConfig() {
```

extension/src/config/config.js
```diff
@@ -1,58 +1,87 @@
...
+    throw new Error(
+      'Please add at least one commercetools project to the config'
+    )
+  if (numberOfAdyenConfigs === 0)
+    throw new Error(
+      'Please add at least one Adyen merchant account to the config'
+    )
+}
+
+loadAndValidateConfig()
```
ahmetoz, line 79: not sure from calling on global scope, should we call it instead of a global call? Are you considering a serverless way 🤔

extension/src/init.js
```diff
@@ -1,22 +1,28 @@
...
+if (envConfig.keepAliveTimeout !== undefined)
+  server.keepAliveTimeout = envConfig.keepAliveTimeout
 server.listen(port, async () => {
-  await ensureResources(ctpClientBuilder.get())
+  const ctpProjectKeys = config.getAllCtpProjectKeys()
+  await pMap(ctpProjectKeys, async (ctpProjectKey) => {
+    const ctpConfig = config.getCtpConfig(ctpProjectKey)
+    if (ctpConfig.ensureResources) {
+      ctpClientBuilder.get(ctpProjectKey)
+      await ensureResources()
```
ahmetoz, line 23: could you put todos in code if it's not done yet ? 🤔 

```suggestion
      await ensureResources(ctpClient, url)
```

extension/src/ctp.js
lojzatran: I've checked your question in the slack channel and no answer yet. I could add a Map here for that purpose, but I'm not sure if it's useful.

extension/src/config/config.js
lojzatran: This will be called anytime this file is being used in the code. So it's sth like a necessary condition for this file to work. It fails really fast if sth is wrong. I don't see any problems here with serverless, what is your concern with serverless?

extension/src/config/config.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/526/commits/3363d8769ca14db7f7fd6e945c7a393d1c846a4d

extension/src/init.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/526/commits/65b4483aec8131465a62654906d3457e95ab4c85

extension/src/config/config.js
ahmetoz: You seem correct, should not be an issue 🤔 👍🏽

extension/src/ctp.js
ahmetoz: That below should be working, with this we don't need to create multiple auth flows for the same project. 
https://github.com/commercetools/nodejs/blob/aa630f042929864d0a06a6461e876ec305d5a775/types/sdk.js#L171-L180

````js
const tokenCache = {
  store: {},
  get(tokenCacheOptions) {
    return this.store[tokenCacheOptions.projectKey]
  },
  set(cache, tokenCacheOptions) {
    this.store[tokenCacheOptions.projectKey] = cache
  }
}
````

then we need to pass the tokenCache to middleware:

````js
const authMiddleware = createAuthMiddlewareForClientCredentialsFlow({
    host: config.ctp.authUrl,
    projectKey,
    credentials: {
      clientId,
      clientSecret,
    },
    fetch,
    tokenCache
  })
````

lojzatran: Thanks for the link. I'm checking it but could not find this option in the docs, maybe it's a hidden feature 😄  https://commercetools.github.io/nodejs/sdk/api/sdkMiddlewareAuth.html. Anyway, a deeper look in the source code shows that if you don't pass anything, there should be a cache by default, see here: https://github.com/commercetools/nodejs/blob/aa630f042929864d0a06a6461e876ec305d5a775/packages/sdk-middleware-auth/src/client-credentials-flow.js#L19. So I guess it's already working like you suggested.

ahmetoz: Don't guess, default impl is here: https://github.com/commercetools/nodejs/blob/a72fd8424250fc50c84197b505a44756ceb407b0/packages/sdk-middleware-auth/src/utils.js#L1-L10

lojzatran: You're right. I had a deeper and still could not understand how it really works if we have multiple clients to cache. I had a deeper look and cannot understand how it works and how we should set it up, so I just added my own cache: https://github.com/commercetools/commercetools-adyen-integration/pull/526/commits/80f8227dda50527fa060fc4aec9aacce1f817b2d

ahmetoz: Why you are not trying this one ? https://github.com/commercetools/commercetools-adyen-integration/pull/526#discussion_r563950126

lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/526/commits/c941656fca7b9e3558a8910b57c9c2493c603939

ahmetoz (approved):

PR #535, opened 2021-01-25
Validate required custom fields

salander85:


extension/src/paymentHandler/payment-handler.js
```diff
@@ -26,6 +26,7 @@ async function handlePayment(paymentObject) {
     return { success: true, data: null }
 
   const paymentValidator = ValidatorBuilder.withPayment(paymentObject)
+    .validateRequiredFields()
```
ahmetoz, line 29: We will pass those to the metadata field of the Adyen payments payload, so I would rename those to 
`validateMetadataFields()` because everything is required for validation logic.

extension/src/validator/validator-builder.js
```diff
@@ -6,6 +6,15 @@ function withPayment(paymentObject) {
   const errors = {}
 
   return {
+    validateRequiredFields() {
+      if (!paymentObject.custom.fields.commercetoolsProjectKey)
```
ahmetoz, line 10: what if `paymentObject.custom` is undefined in the request, it would cause NPE, maybe check other similar validation method calls.

Also what if the value is there but it has whiteSpace chars? Should it be a warning too?

salander85: Good point 👍 I think for whiteSpace it should also fail...

salander85: c0fc985

extension/src/paymentHandler/payment-handler.js
salander85: c0fc985

extension/src/validator/validator-builder.js
ahmetoz: please validate all in once not step by step 😉 

with looking at the code, it also might be there is no way to put undefined or null into the paymentObject.custom 🤔 or even to fields could have some field if they are empty (not sure) maybe you need to check what API returns.

Also, whiteSpace check should not be checked that way, those following texts are also not valid: "projectName ", " projectName", "project name". maybe check if the string has an empty string?

I'm also not sure from that please check both ct and adyen docs to ensure the possible formats, the benefit is failing fast that way so we could avoid extra calls and exception handling in the extension module.

extension/src/validator/validator-builder.js
```diff
@@ -6,8 +6,21 @@ function withPayment(paymentObject) {
...
+            errorMessages.MISSING_REQUIRED_FIELDS_CTP_PROJECT_KEY
+      if (!paymentObject.custom.fields.adyenMerchantAccount ||
+          paymentObject.custom.fields.adyenMerchantAccount === ' ')
+        errors.missingRequiredAdyenMerchantAcc =
+            errorMessages.MISSING_REQUIRED_FIELDS_ADYEN_MERCHANT_ACCOUNT
+      return this
+    },
     validateRequestFields() {
-      if (!paymentObject.custom) return this
+      if (!paymentObject.custom || errors.missingRequiredCtpProjectKey || errors.missingRequiredAdyenMerchantAcc)
```
lojzatran, line 22: I don't think this method needs to check `missingRequiredCtpProjectKey` and `missingRequiredAdyenMerchantAcc`. It checks only fields with `Request` ending, so it should not care about other fields.

extension/src/validator/validator-builder.js
```diff
@@ -6,8 +6,21 @@ function withPayment(paymentObject) {
   const errors = {}
 
   return {
+    validateMetadataFields() {
+      if (!paymentObject.custom && !paymentObject.custom.fields) return this
+      if (!paymentObject.custom.fields.commercetoolsProjectKey ||
+          paymentObject.custom.fields.commercetoolsProjectKey === ' ')
```
lojzatran, line 12:
```suggestion
          paymentObject.custom.fields.commercetoolsProjectKey.trim() === '')
```

extension/src/validator/validator-builder.js
```diff
@@ -6,8 +6,21 @@ function withPayment(paymentObject) {
...
+        errors.missingRequiredCtpProjectKey =
+            errorMessages.MISSING_REQUIRED_FIELDS_CTP_PROJECT_KEY
+      if (!paymentObject.custom.fields.adyenMerchantAccount ||
+          paymentObject.custom.fields.adyenMerchantAccount === ' ')
```
lojzatran, line 16:
```suggestion
          paymentObject.custom.fields.adyenMerchantAccount.trim() === '')
```

lojzatran (changes requested):

extension/src/validator/validator-builder.js
salander85: So, paymentDraft.custom is optional - so it could be undefined or null. If custom is provided but not fields API adds empty fields.

salander85: Quickly checked the docs; could not find anything about the formats other than both CT and adyen are generating these values - as they need to have url conformity i assume whitespaces are not allowed, lead to errors.

extension/src/validator/validator-builder.js
salander85: Changed to check if the field contains any whitespace 9f61a99

extension/src/validator/validator-builder.js
salander85: 9f61a99

extension/src/validator/validator-builder.js
salander85: 9f61a99

extension/src/validator/validator-builder.js
salander85: Changed to check if the field contains any whitespace 9f61a99

ahmetoz (approved):
good work 👍🏽

extension/src/validator/validator-builder.js
lojzatran: Actually this is the requirement for the project key in CT: `May only contain between 2 and 36 alphanumeric characters, underscores, or hyphens (no white space or special characters like ñ, ü, #, %).`

lojzatran (approved):

PR #541, opened 2021-01-29
notification module - support multitenancy

lojzatran:
I added support for multitenancy for notification module in this PR. 

Additionally I also did small changes for the extension module so that the common files in both extension and notification modules are more or less the same. This should help us later when we move them to a common shared module.


extension/src/init.js
```diff
@@ -16,13 +16,17 @@ if (envConfig.keepAliveTimeout !== undefined)
...
-      await ensureResources(ctpClient)
-    }
-  })
+  await pMap(
+    ctpProjectKeys,
+    async (ctpProjectKey) => {
+      const ctpConfig = config.getCtpConfig(ctpProjectKey)
+      if (ctpConfig.ensureResources) {
+        const ctpClient = ctpClientBuilder.get(ctpConfig)
+        await ensureResources(ctpClient)
```
ahmetoz, line 25: I've changed this in multitenancy-extension-IT branch, also envConfig to moduleConfig..
```suggestion
await ensureResources(ctpClient, moduleConfig.apiExtensionBaseUrl)
```

extension/src/init.js
```diff
@@ -16,13 +16,17 @@ if (envConfig.keepAliveTimeout !== undefined)
...
+      }
+    },
+    { concurrency: 5 }
```
ahmetoz, line 28: What is the benefit of limiting concurrency here?

notification/package.json
```diff
@@ -72,6 +72,7 @@
     "lint-staged": "^10.5.3",
     "mocha": "^8.2.1",
     "nyc": "^15.1.0",
+    "p-map": "^4.0.0",
```
ahmetoz, line 75: how was it working without this?

extension/src/service/web-component-service.js
```diff
@@ -13,33 +13,48 @@ function getPaymentMethods(merchantAccount, getPaymentMethodsRequestObj) {
...
 }
 
-function makePayment(merchantAccount, makePaymentRequestObj) {
+function makePayment(
+  merchantAccount,
+  commercetoolsProjectKey,
+  makePaymentRequestObj
+) {
   const adyenCredentials = config.getAdyenConfig(merchantAccount)
+  extendRequestObjWithMetadata(makePaymentRequestObj, commercetoolsProjectKey)
```
ahmetoz, line 22: web-component-service.js getting uglier with every change, maybe it's to time to restructure it? Let me know if you want to discuss making it cleaner. Might be another PR for it?

ahmetoz: Not in diff but looks like an issue in the validation logic, this should be an invalid request if custom.fields.merchantAccount and request.merchantAccount, right? 
```
if (!requestObj.merchantAccount)
    requestObj.merchantAccount = adyenMerchantAccount
```

extension/test/cleanup-ctp-project.js
```diff
@@ -4,8 +4,9 @@ const config = require('../src/config/config')
 
 before(async () => {
   const ctpProjectKeys = config.getAllCtpProjectKeys()
-  for (const projectKey of ctpProjectKeys) {
-    const ctpClient = ctpClientBuilder.get(projectKey)
+  for (const ctpProjectKey of ctpProjectKeys) {
+    const ctpConfig = config.getCtpConfig(ctpProjectKey)
```
ahmetoz, line 8: I did not get those cascading changes on every test, I guess it was better before in extension because when you know the project key you know the ctp config ? 🤔 Am I missing something ?

notification/src/api/notification/notification.controller.js
```diff
@@ -1,30 +1,97 @@
...
+  constructor({ stack, message, notification, isRecoverable }) {
+    super()
+    this.stack = stack
+    this.message = message
+    this.notification = JSON.stringify(notification)
+    this.isRecoverable = isRecoverable
+  }
+}
 
 // TODO: add JSON schema validation:
```
ahmetoz, line 19: Please remove obsolete TODOS.

notification/src/api/notification/notification.controller.js
```diff
@@ -1,30 +1,97 @@
...
+      )
+    }
     return sendAcceptedResponse(response)
   } catch (err) {
     logger.error(
       { adyenRequestBody: `${body}`, err },
       'Unexpected exception occurred.'
     )
-    return httpUtils.sendResponse(response, 500)
+    if (err.isRecoverable) {
```
ahmetoz, line 42: What is the impact on Adyen's side, should we return 500 ? or accepted for every case? 
When Adyen stops sending notifications?

notification/src/config/init/ensure-interface-interaction-custom-type.js
```diff
@@ -1,11 +1,24 @@
...
+  const commercetoolsProjectKeys = config.getAllCtpProjectKeys()
+  for (const commercetoolsProjectKey of commercetoolsProjectKeys) {
+    const ctpConfig = config.getCtpConfig(commercetoolsProjectKey)
+    if (ctpConfig.ensureResources)
+      await ensureInterfaceInteractionCustomType(ctpConfig)
+  }
+}
+
+async function ensureInterfaceInteractionCustomType(ctpProjectConfig) {
+  const logger = mainLogger.child({
```
ahmetoz, line 17: Do you want to add this logger context to the extension module?  😅

notification/src/api/notification/notification.controller.js
```diff
@@ -1,30 +1,97 @@
...
 const logger = require('../../utils/logger').getLogger()
 
-const ctpClient = ctp.get(config)
+class ValidationError extends Error {
+  constructor({ stack, message, notification, isRecoverable }) {
+    super()
+    this.stack = stack
+    this.message = message
+    this.notification = JSON.stringify(notification)
+    this.isRecoverable = isRecoverable
```
ahmetoz, line 15: should we manage code flow with exceptions? It's not suggested most of time because errors are expensive as they collects all stack in memory (not sure in nodejs)? Is there any other way ?

notification/src/handler/notification/notification.handler.js
```diff
@@ -154,6 +159,34 @@ function calculateUpdateActionsForPayment(payment, notification) {
...
+      'Wrong transaction state passed. ' +
+        `currentState: ${currentState}, newState: ${newState}`
+    )
+
+  return transactionStateFlow[newState] - transactionStateFlow[currentState]
+}
+
 function getAddInterfaceInteractionUpdateAction(notification) {
   // strip away sensitive data
   delete notification.additionalData
```
ahmetoz, line 192: maybe we should not remove all data as metadata might be good to have for tracability also modification.action looks like good to have ?

ahmetoz (changes requested):
Do we have an integration test case scenario like: 
- 2 projects and 2 parallel calls to the module, and ensuring which project is updated?

extension/src/init.js
lojzatran: Benefit is in case extension has hundreds of projects at once, then this module will call only 5 projects at once to ensure resources.

notification/package.json
lojzatran: Good question, actually I don't know 😕  It's used in the tests, I would expect them to fail.

notification/src/api/notification/notification.controller.js
lojzatran: If we accept 200, they will stop sending notifications. I was thinking that in case notification doesn't have a project configured, it should return 500 so that Adyen will keep on retry. It could be that one forget to configure, so at least this will give them a chance to reconfigure and fix that.

notification/src/config/init/ensure-interface-interaction-custom-type.js
lojzatran: Thanks, I forgot about it 😅  I will add it.

notification/src/handler/notification/notification.handler.js
lojzatran: We could, but it means we would need to check what data could be passed in `additionalData` and what they can contain. And we need to verify it. And it could have lots of possible fields inside: https://docs.adyen.com/development-resources/webhooks/additional-settings.

extension/src/service/web-component-service.js
lojzatran: > web-component-service.js getting uglier with every change, maybe it's to time to restructure it? Let me know if you want to discuss making it cleaner. Might be another PR for it?

We can discuss it, but I'm not sure when and how. This extending thing is just for some requests, so it's not very easy to generalize.

lojzatran: > Not in diff but looks like an issue in the validation logic, this should be an invalid request if custom.fields.merchantAccount and request.merchantAccount, right?
> 
> ```
> if (!requestObj.merchantAccount)
>     requestObj.merchantAccount = adyenMerchantAccount
> ```

I tried to remove it, but for example for Cancellation or Refund this is valid code. For them we create a request from a transaction and in transaction we don't have `merchantAccount`. One way is to add this part only in the methods that use it. Then it will cause less confusion, but we have some code duplication. WDYT is better?

extension/test/cleanup-ctp-project.js
lojzatran: Yes, that's correct. Now there could be multiple ctp projects, so I made this kind of clean up.

notification/src/api/notification/notification.controller.js
lojzatran: I also read that exception is more expensive, but I don't think it makes any noticeable difference here, especially when we're dealing with network calls anyway and this app is not super high load app. So I had a look at another aspect of the problem and [in this answer](https://stackoverflow.com/a/1504400) a guy wrote:

```
Personally I like throwing Exceptions for business rule validation (not so much for user input validation) because it forces the problem to be handled upstream. If my business objects returned some kind of validation result, it could be ignored by the caller. Call me a cowboy if you wish :)
```

Since these validations are not user input validations, but misconfiguration of the app or sth wrong with Adyen's side, throwing exception is ok. Validation is also ok, but I feel like when sth is misconfigured or an expected format of notification is not coming, then it's more like an exception than validation.

extension/src/init.js
ahmetoz: It could stay but imo requests are not going to the same server or same ctp client, so putting a concurrency for it does not make sense in terms of the node's working model. 

Also, do you know what could happen if ensure resources fails for a project but then stops? In the next run, it should not create resources on those projects, right?

notification/package.json
ahmetoz: Looks like merge conflicts issues because its set on the master branch: https://github.com/commercetools/commercetools-adyen-integration/blob/master/notification/package.json#L55

extension/src/service/web-component-service.js
ahmetoz: > We can discuss it, but I'm not sure when and how. This extending thing is just for some requests, so it's not very easy to generalize.

Let me check it, we could discuss this in a different PR.. For now, pass my review comment.

ahmetoz: My concerns not for the place that we assign merchant account to request object (i.e: makePayment, cancel and refund request), here when custom.fields.adyenMerchantAccount is different than requestObj.merchantAccount we should not allow sending it to adyen because we can not match notification for it.. I wonder if we could add this validation before sending a request.

extension/test/cleanup-ctp-project.js
ahmetoz: So I'm not missing anything, and it was cleaner before ?

This part is added everywhere on client creation on tests : `const ctpConfig = config.getCtpConfig(ctpProjectKey)` 
this was in ctpClient get function before. imo you could get all config with only knowing ctpProjectKey (before it was that way).

Anyways you could keep it if you see a benefit.

lojzatran: Yes, you need to know the project key now to get the config. Before it was not necessary as you have only one config.

notification/src/api/notification/notification.controller.js
ahmetoz: I feel that we need to consult Adyen for best practices here, can you ask them ?

So in the basics when the module acknowledges the notification, Adyen servers will not send it back. 
In docs says: https://docs.adyen.com/development-resources/webhooks#accept-notifications
if the Adyen server is not receiving `[accepted]` from the integration it will block all notifications (looks like queued)
https://docs.adyen.com/development-resources/webhooks/best-practices#queued-notifications so is it good to cascade notifications on the queue until the devs configured it well (does not feel right on the production use case).

lojzatran: I will ask them but I remember we asked them many times in the past and they always say accept it and it was not very helpful 😄

notification/src/api/notification/notification.controller.js
ahmetoz: Here the client is Adyen and there is no way to be handled upstream.

Also, there is no challenge on the business calls or some cases like injection in different methods needed and you are throwing an exception to handle it, which exceptions should not be used for that purpose. 

I feel here it's not that hard to check the notification payload is valid or not. Logic is similar to what we are doing on the extension module's validations.

The only place we needed to use exception to check if the payload is a valid JSON 😉  

Note: I won't block the review because of this, you might skip my comment.

notification/src/config/init/ensure-interface-interaction-custom-type.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/4d0741f4b2d6113bda662ffdf5d7209d4d4faa27

extension/test/cleanup-ctp-project.js
ahmetoz: I feel these discussions will not end 🚀 

Why you are not passing only the projectKey then, because you could get the ctpConfig inside the ctp.js ? Also it's not in this branch that we are passing projectKey to the ctp.js.

extension/test/e2e/klarna.spec.js
```diff
@@ -70,7 +71,7 @@ describe('::klarnaPayment::', () => {
   it(
     'when payment method is klarna and process is done correctly, ' +
       'then it should successfully finish the payment',
-    async function () {
+    async function func() {
```
JudeNiroshan, line 73:
```suggestion
    async () => {
```

notification/src/api/notification/notification.controller.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/840b242977e2cb06225f8ebaa881e269820122a4

extension/test/cleanup-ctp-project.js
lojzatran: I'm not passing just projectKey because I felt like the client should not load config, but rather it should get the config from outside. It's also easier for testing if we can pass modified config.

extension/test/integration/make-payment.handler.spec.js
```diff
@@ -1,97 +1,122 @@
 const { expect } = require('chai')
 
 const ctpClientBuilder = require('../../src/ctp')
-const iTSetUp = require('./integration-test-set-up')
-const constants = require('../../src/config/constants')
 const config = require('../../src/config/config')
+const constants = require('../../src/config/constants')
+const iTSetUp = require('./integration-test-set-up')
+
+describe('::make-payment with multiple adyen accounts use case::', () => {
```
JudeNiroshan, line 8: Shouldn't we also add a scenario of with multiple CTP projects but one has incorrect credentials?

extension/src/init.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/4d0741f4b2d6113bda662ffdf5d7209d4d4faa27

notification/src/config/config.js
```diff
@@ -1,81 +1,109 @@
...
-        'The "ADYEN_SECRET_HMAC_KEY" environment variable is missing to be able to verify notifications, ' +
-          'please generate a secret HMAC key in Adyen Customer Area ' +
-          'or set "ADYEN_ENABLE_HMAC_SIGNATURE=false" to disable the verification feature.'
+        `[${adyenMerchantAccount}]: The "secretHmacKey" config variable is missing to be able to verify ` +
+          `notifications, please generate a secret HMAC key in Adyen Customer Area ` +
+          `or set "enableHmacSignature=false" to disable the verification feature.`
       )
   }
+}
+
```
JudeNiroshan, line 100: It would be interesting to see how many ctpProjects and adyenProjects are configured whenever I start up the notification server as a server log message

JudeNiroshan (changes requested):

extension/test/integration/make-payment.handler.spec.js
ahmetoz: using integration tests for obvious failures has no benefit in general, we might have a unit test for such cases? Both in config and validation cases are unit tests. @lojzatran could you double check.

lojzatran:
> Do we have an integration test case scenario like:
> 
> * 2 projects and 2 parallel calls to the module, and ensuring which project is updated?

https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/b5cb18f866f218383ec823196472f8e86564b0f0

extension/test/integration/make-payment.handler.spec.js
lojzatran: I don't think it's necessary to use integration tests for error cases. We already have some unit tests here.

extension/test/e2e/klarna.spec.js
lojzatran: Hey Jude, in this method I cannot change it back to lambda function as I have `this.timeout(60000)` inside this method (one line below). Additionally I cannot have anonymous function as eslint is complaining there. Thus it must be a function and also has a name `func`.

notification/src/config/config.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/9c0e71db6c64a41aab143e72c5ab26f66f0687ca

lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/d8037b2c2786c77d76a52cf25b629f60d87112e0

extension/test/cleanup-ctp-project.js
```diff
@@ -4,8 +4,9 @@ const config = require('../src/config/config')
 
 before(async () => {
   const ctpProjectKeys = config.getAllCtpProjectKeys()
-  for (const projectKey of ctpProjectKeys) {
-    const ctpClient = ctpClientBuilder.get(projectKey)
+  for (const ctpProjectKey of ctpProjectKeys) {
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/a7d3cbde1a1bdb31657a42d7c26739f0e8d4a9dd

extension/src/service/web-component-service.js
```diff
@@ -48,8 +63,13 @@ function manualCapture(merchantAccount, manualCaptureRequestObj) {
   )
 }
 
-function cancelPayment(merchantAccount, cancelRequestObj) {
+function cancelPayment(
+  merchantAccount,
+  commercetoolsProjectKey,
+  cancelRequestObj
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/6ec1dd1df29af21bd458047fb8e831bf742657a2

extension/test/cleanup-ctp-project.js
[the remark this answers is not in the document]
lojzatran: I removed pMap as it's not very useful in our case: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/a2fda293e4e3dbf23d0b28ec262f3bea95a07d69

extension/src/init.js
lojzatran: I removed pMap as it's not very useful for our usage anyway: https://github.com/commercetools/commercetools-adyen-integration/pull/541/commits/a2fda293e4e3dbf23d0b28ec262f3bea95a07d69

If it fails for a project, it should fail the whole extension. During the next run, it first checks if resources are there. If not, it will create, if resources are there, it will not create again.

extension/src/service/web-component-service.js
lojzatran: But if it's different, Adyen will reject such request. We use `custom.fields.adyenMerchantAccount` to get an adyen config and then we send the request to it. If request contains a merchantAccount that does not exist, it will fail immediately and it will not go to notification.

JudeNiroshan (approved):

extension/src/service/web-component-service.js
ahmetoz: for instance:  is it possible to send a payment for the `merchant2` when the `custom.fields.adyenMerchantAccount` is set to `merchant1`, in case both `merchant1` and `merchant2` is valid and configured?

lojzatran: It's possible to send. But Adyen will reject it I guess. Let's take it step by step:

1. `custom.fields.adyenMerchantAccount` is used to load the adyen config from the configuration. In this case we will load configuration for `merchant1`
2. In the request, we have `merchant2` set as `merchantAccount` and we will send this `merchant1` account.

So I hope in this case adyen will reject it :) But it's true we did not test this case as we don't have 2 separate accounts. Our "accounts" for testing have the same clientKey and apiKey, so I guess it's just a namespace in the same adyen account. We can test it when we get real 2 accounts from Adyen.

ahmetoz (approved):

extension/src/service/web-component-service.js
ahmetoz: For me, this looks like a validation issue when you have the same API client for different merchant accounts (which is possible by Adyen) and the module should not allow the payments when there is a difference in the metadata and make payment.. but will not block the PR just for this, as you suggested please handle the testing when you have the accounts.

PR #560, opened 2021-02-12
Fix notification functions and refactorings.

ahmetoz:


notification/src/lambda.js
```diff
@@ -20,37 +16,27 @@ exports.handler = async function (event) {
...
-        notification.NotificationRequestItem.merchantAccountCode
-      const ctpProjectConfig = config.getCtpConfig(commercetoolsProjectKey)
-      const adyenConfig = config.getAdyenConfig(adyenMerchantAccount)
+  try {
+    for (const notification of notificationItems) {
+      const ctpProjectConfig = getCtpProjectConfig(notification)
+      const adyenConfig = getAdyenConfig(notification)
 
-      if (!initialised) {
-        await setup.ensureInterfaceInteractionCustomType(ctpProjectConfig)
```
lojzatran, line 35: Why ensure is removed for lambda?

ahmetoz: I thought it's better if we have consistency between modules and docs, in extension functions are not creating resources (also not in GCP function of notification). 

I see it could be done in service but not really sure if it's the best practice to keep the `state` in a function for the creation of the resource as it's a one-time thing needed on setup.

lojzatran (approved):

PR #561, opened 2021-02-12
Support sync for resources.

ahmetoz:


extension/test/unit/config/init/ensure-resources.spec.js
```diff
@@ -61,27 +61,26 @@ describe('Ensure resources', () => {
     )
   })
 
-  it('should skip creating resources when they already exist', async () => {
```
lojzatran, line 64: Why did you remove this test?

lojzatran (approved):

extension/test/unit/config/init/ensure-resources.spec.js
ahmetoz: it will not skip anymore instead it will check the diff and create actions, I tried to use but those tests are mostly based on mocking the API calls and error-prone, I wrote an integration test instead, unit tests might check error handling logic.

PR #564, opened 2021-02-15
Adapt lambda functions to multitenancy flow.

ahmetoz:
Developed together with @salander85


deployment-examples/aws/serverless/extension/README.md
```diff
@@ -1,5 +1,14 @@
...
-2. When the extension module is run as AWS Lambda **it will NOT create** the required resources like custom types or commercetools API extension for you. 
+1. Run 'npm run zip-lambda-function'. It helps to execute following steps
+   - Rename index.lambda.js to index.js
+   - For deployment to AWS Lambda, bundle the following files and folders into a .zip file and upload it in AWS console.
+      - src
+      - node_modules
+      - resources
+      - index.js
+      - package.json
+2. Specify `index.handler` as the entry point for the AWS Lambda function.
```
leungkinghin, line 11: Has it been tested in AWS for this naming already?
For the GCP case, it is specified as 'extensionTrigger' and it doesn't require to be named as 'index.extensionTrigger'.

notification/index.lambda.js
```diff
@@ -1,10 +1,11 @@
...
-const { getNotificationForTracking } = require('./utils/commons')
-const { getCtpProjectConfig, getAdyenConfig } = require('./utils/parser')
+const handler = require('./src/handler/notification/notification.handler')
+const logger = require('./src/utils/logger').getLogger()
+const { getNotificationForTracking } = require('./src/utils/commons')
+const { getCtpProjectConfig, getAdyenConfig } = require('./src/utils/parser')
 
 exports.handler = async (event) => {
-  const { notificationItems } = event
+  const body = event.body ? JSON.parse(event.body) : event
```
leungkinghin, line 7: Shall we update corresponding unit test case ? so that the mocking event in unit test contains body attribute.

extension/index.lambda.js
```diff
@@ -1,11 +1,12 @@
...
-const paymentHandler = require('../src/paymentHandler/payment-handler')
+const utils = require('./src/utils')
+const paymentHandler = require('./src/paymentHandler/payment-handler')
 
 const logger = utils.getLogger()
 
 exports.handler = async (event) => {
   try {
-    const paymentResult = await paymentHandler.handlePayment(event.resource.obj)
+    const body = event.body ? JSON.parse(event.body) : event
```
leungkinghin, line 8: Same as https://github.com/commercetools/commercetools-adyen-integration/pull/564/files#r576360178

notification/index.lambda.js
ahmetoz: > there is a body attribute always inside event object. Shall we update corresponding unit test case as well ?
We were not sure about it and thought it could be related to the API gateway, can you ensure it ? If yes we could update it, for now we are not sure.

deployment-examples/aws/serverless/extension/README.md
ahmetoz: It's different than GCP:
Screenshot 2021-02-15 at 19 55 17

notification/index.lambda.js
ahmetoz: here: https://stackoverflow.com/a/41656022

leungkinghin: That means we have two options when establishing API gateway on lambda. I think it would be fine to keep unit test unchanged at this moment which pretends request from lambda integration, since we cannot 100% sure the json format for lambda proxy integration until we adapt it.

leungkinghin (approved):

deployment-examples/aws/serverless/extension/README.md
```diff
@@ -1,5 +1,14 @@
...
+ > The default value in the console is index.handler which calls exports.handler in index.js
+3. When the extension module is run as AWS Lambda **it will NOT create** the required resources like custom types or commercetools API extension for you.
```
lojzatran, line 13: Now we removed the ensure thing, should we rather say they should use the script?

ahmetoz: Good point, this document is not the latest version of this it's not only commercetools resources I simply started with:

> In order to setup the extension module config, follow our how to run guide.

https://github.com/commercetools/commercetools-adyen-integration/tree/disable-ensure-resources/deployment-examples/aws/serverless/extension#deploy-as-aws-lambda

lojzatran (approved):

notification/index.lambda.js
butenkor: The check seems obsolete/not obvious. Why we do not add to such places some comments or links?

PR #621, opened 2021-03-15
Update HowToRun.md

leungkinghin:
Format the table in HowToRun.md for newly added authentication setting in extension configuration


butenkor:
@leungkinghin Did you format it or broke existing formatting - at least from the perspective of human-readability? :)

leungkinghin:
@butenkor It is generated by the prettier formatter forcefully, otherwise the style checking would fail the build :(

Issue #602, opened 2021-03-03
Faulty try/catch in notification processing


GergelyToth:
https://github.com/commercetools/commercetools-adyen-integration/blob/a63624fb94d294f4ec50ae99e06a142e922c2ed3/notification/src/utils/parser.js#L15-L20

If the `metadata.commercetoolsProjectKey` is not present in the `additionalData`, but the property itself exists in the notification, this try/catch will not fail, and instead will pass `undefined` as project key, which will throw the following error:

```
Configuration is not provided. Please update the configuration. ctpProjectKey: [undefined]
```

lojzatran:
Hi @GergelyToth,

Thank you for reporting this potential issue. I had a look at the code and here's what I found out:

The notification module requires `metadata.commercetoolsProjectKey` to work correctly as it needs to save the notification to the matching payment in the project. Without `commercetoolsProjectKey` it cannot determine the CTP project. Therefore we expect to have `metadata.commercetoolsProjectKey` in every notification.

Can you write in which type of notification you get the problem and also post the notification here so we can investigate further?

Best regards,
Lam Tran

GergelyToth:
Hi Lam,

I have not been able to successfully retrieve the full notification from the error message that was throw. 
The integration is successfully working on Apple Pay, Google Pay and card payments, but it fails on Paypal authorization notifications in our end, and unfortunately, even after contacting Adyen to request the full notification, they could not provide the necessary information.

Kind regards,
Greg

butenkor:
@GergelyToth I had a talk with Adyen and got the following reply:
```
At time of sending the notification we have for some local payment methods not the required info in that specific system.
There is a setting “skip sending on PAL” that fixes the issue for these local payment methods. 
Support can enable that setting on the notification endpoint of the merchant. 

If you/they could please request this setting on the notification endpoint it would be great. 
I now don’t have the right context of the merchant (accounts, etc.)
```
Let us know if this helped.

GergelyToth:
Hi,

The example notification, fired from the Adyen server notification backend is failing for the reason mentioned by my first comment. The try/catch validation check doesn't catch the project key missing, and is treating the notification as it something still recoverable, and should be processed by it, and will pass the `undefined` key down, where it fails on: https://github.com/commercetools/commercetools-adyen-integration/blob/a63624fb94d294f4ec50ae99e06a142e922c2ed3/notification/src/config/config.js#L17

Please find the example notification and the server response below:

``json
{
[...]

PR #629, opened 2021-03-23
Fix "metadata.commercetoolsProjectKey" check and also return "accepted" by default.

ahmetoz:
Addresses: #602

### DIFF
--- a/notification/index.googleFunction.js
+++ b/notification/index.googleFunction.js
@@ -24,8 +24,8 @@ exports.notificationTrigger = async (request, response) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (err.isRecoverable) {
-      return response.status(400).send(err.message)
+    if (!err.isRecoverable) {
+      return response.status(500).send(err.message)
     }
   }
 
--- a/notification/index.googleFunction.js
+++ b/notification/index.googleFunction.js
@@ -24,7 +24,7 @@ exports.notificationTrigger = async (request, response) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (!err.isRecoverable) {
+    if (err.isRecoverable) {
       return response.status(500).send(err.message)
     }
   }
--- a/notification/index.googleFunction.js
+++ b/notification/index.googleFunction.js
@@ -24,7 +24,7 @@ exports.notificationTrigger = async (request, response) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (err.isRecoverable) {
+    if (err.retry) {
       return response.status(500).send(err.message)
     }
   }
--- a/notification/index.lambda.js
+++ b/notification/index.lambda.js
@@ -4,6 +4,7 @@ const { getNotificationForTracking } = require('./src/utils/commons')
 const { getCtpProjectConfig, getAdyenConfig } = require('./src/utils/parser')
 
 exports.handler = async (event) => {
+  // Reason for this check: event.body is a string of payload when AWS API Gateway is used.
   const body = event.body ? JSON.parse(event.body) : event
   const { notificationItems } = body
   if (!notificationItems) {
@@ -33,7 +34,7 @@ exports.handler = async (event) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (err.isRecoverable) {
+    if (!err.isRecoverable) {
       throw err
     }
   }
--- a/notification/index.lambda.js
+++ b/notification/index.lambda.js
@@ -34,7 +34,7 @@ exports.handler = async (event) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (err.isRecoverable) {
+    if (err.retry) {
       throw err
     }
   }
--- a/notification/index.lambda.js
+++ b/notification/index.lambda.js
@@ -34,7 +34,7 @@ exports.handler = async (event) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (!err.isRecoverable) {
+    if (err.isRecoverable) {
       throw err
     }
   }
--- a/notification/index.lambda.js
+++ b/notification/index.lambda.js
@@ -4,7 +4,7 @@ const { getNotificationForTracking } = require('./src/utils/commons')
 const { getCtpProjectConfig, getAdyenConfig } = require('./src/utils/parser')
 
 exports.handler = async (event) => {
-  // Reason for this check: event.body is a string of payload when AWS API Gateway is used.
+  // Reason for this check: if AWS API Gateway is used then event.body is provided as a string payload.
   const body = event.body ? JSON.parse(event.body) : event
   const { notificationItems } = body
   if (!notificationItems) {
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -11,7 +11,7 @@ class CommercetoolsError extends Error {
     super()
     this.stack = stack
     this.message = message
-    this.isRecoverable = this._isRecoverable(statusCode);
+    this.isRecoverable = this._isRecoverable(statusCode)
   }
 
   /**
@@ -22,10 +22,12 @@ class CommercetoolsError extends Error {
    * If commercetools status code is not OK but also not 5xx or 409 then return `accepted` -> non recoverable
    */
   _isRecoverable(statusCode) {
-      return !(statusCode !== undefined &&
-        statusCode > 399 &&
-        statusCode !== 409 &&
-        statusCode < 500)
+    return !(
+      statusCode !== undefined &&
+      statusCode > 399 &&
+      statusCode !== 409 &&
+      statusCode < 500
+    )
   }
 }
 
@@ -131,7 +133,7 @@ async function updatePaymentWithRepeater(
             ` max retries. Failed actions: ${JSON.stringify(
               _obfuscateNotificationInfoFromActionFields(updateActions)
             )}`,
-          statusCode: err.statusCode
+          statusCode: err.statusCode,
         })
       }
       /* eslint-disable-next-line no-await-in-loop */
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -214,9 +214,7 @@ function compareTransactionStates(currentState, newState) {
     !transactionStateFlow.hasOwnProperty(newState)
   )
     throw new CommercetoolsError({
-      message:
-        'Wrong transaction state passed. ' +
-        `currentState: ${currentState}, newState: ${newState}`,
+      message: `Wrong transaction state passed. CurrentState: ${currentState}, newState: ${newState}`,
     })
 
   return transactionStateFlow[newState] - transactionStateFlow[currentState]
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -15,8 +15,8 @@ class CommercetoolsError extends Error {
   }
 
   /**
-   * recoverable: notification delivery can be retried by Adyen
-   * non recoverable: notification delivery can not be retried by Adyen as it most probably would fail again
+   * recoverable: notification delivery can be retried by Adyen (return 500)
+   * non recoverable: notification delivery can not be retried by Adyen as it most probably would fail again (return "accepted")
    * If commercetools status code is defined and is 5xx then return `500` to Adyen -> recoverable
    * If during communication with commercetools we got a `NetworkError` then return `500` -> recoverable
    * If commercetools status code is not OK but also not 5xx or 409 then return `accepted` -> non recoverable
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -106,8 +106,8 @@ async function updatePaymentWithRepeater(
         throw new CommercetoolsError({
           stack: err.stack,
           message:
-            `Unexpected error during updating a payment with ID: ${currentPayment.id}. Exiting. ` +
-            `Actions: ${JSON.stringify(
+            `Unexpected error on payment update with ID: ${currentPayment.id}. Exiting. ` +
+            `Failed actions: ${JSON.stringify(
               _obfuscateNotificationInfoFromActionFields(updateActions)
             )}`,
           statusCode: err.statusCode,
@@ -124,7 +124,7 @@ async function updatePaymentWithRepeater(
           message:
             `${retryMessage} Won't retry again` +
             ` because of a reached limit ${maxRetry}` +
-            ` max retries. Actions: ${JSON.stringify(
+            ` max retries. Failed actions: ${JSON.stringify(
               _obfuscateNotificationInfoFromActionFields(updateActions)
             )}`,
         })
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -141,8 +141,6 @@ async function updatePaymentWithRepeater(
 }
 
 function _obfuscateNotificationInfoFromActionFields(updateActions) {
-  if (!updateActions) return updateActions
-
   const copyOfUpdateActions = _.cloneDeep(updateActions)
   copyOfUpdateActions
     .filter((value) => value.action === 'addInterfaceInteraction')
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -11,9 +11,17 @@ class CommercetoolsError extends Error {
     super()
     this.stack = stack
     this.message = message
-    // statusCode === 0 is NetworkError, default is recoverable.
+
+    /*
+      We do not want to block notifications coming by Adyen.
+      So with this error it will be accepted by notification module by default.
+
+      Only we do not accept for statusCode === 0 (NetworkError) or 5xx errors like 500.
+      Note that, our client configured to do retries for statusCode=503 cases.
+      So when isRecoverable assigned to true the notification will not respond with accepted but with 500 status code.
+    */
     this.isRecoverable =
-      statusCode !== undefined ? statusCode !== 0 && statusCode < 500 : true
+      statusCode !== undefined ? statusCode === 0 && statusCode >= 500 : false
   }
 }
 
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -23,12 +23,7 @@ class CommercetoolsError extends Error {
    * If commercetools status code is not OK but also not 5xx or 409 then return `accepted` -> non recoverable
    */
   _shouldRetry(statusCode) {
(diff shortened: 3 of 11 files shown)


notification/index.googleFunction.js
```diff
@@ -24,8 +24,8 @@ exports.notificationTrigger = async (request, response) => {
       { notification: getNotificationForTracking(notificationItems), err },
       'Unexpected exception occurred.'
     )
-    if (err.isRecoverable) {
-      return response.status(400).send(err.message)
+    if (!err.isRecoverable) {
+      return response.status(500).send(err.message)
```
butenkor, line 28: Did not we say that we return 500 on recoverable errors only? On non recoverable it has to be `accepted` so that we do not block other notifications.

ahmetoz: With this implementation, I changed the meaning of recoverable, if it’s recoverable it will be accepted, if not it will be rejected. Was it better before ? 

In this case: 
- All validation errors will be `err.isRecoverable = true` means will return "accepted" to Adyen service.
- Unexpected errors like bugs will be `err.isRecoverable = undefined` means will return 500 to Adyen service.
- All `commercetools` related API errors like 400 will be `err.isRecoverable=true` means will return "accepted" to Adyen service.
- Errors like NetworkError (statusCode=0) or API errors like statusCode=5xx will be `err.isRecoverable ` means will return 500 to Adyen service.

lojzatran (changes requested):

JudeNiroshan (changes requested):

notification/index.googleFunction.js
butenkor: To be honest I find it confusing.

> All validation errors will be err.isRecoverable = true means will return "accepted" to Adyen service.

In this case from integration perspective it is not recoverable but because we do not want to block we return OK/accepted. It think previously it was easier to understand the intention of recoverable vs. non recoverable.

> Unexpected errors like bugs will be err.isRecoverable = undefined means will return 500 to Adyen service

Unexpected errors should lead to accepted and not 500, right?

> All commercetools related API errors like 400 will be err.isRecoverable=true means will return "accepted" to Adyen service.

Again, it is confusing as from integration perspective we have an obviously not recoverable 400 error.

notification/index.lambda.js
```diff
@@ -4,6 +4,7 @@ const { getNotificationForTracking } = require('./src/utils/commons')
 const { getCtpProjectConfig, getAdyenConfig } = require('./src/utils/parser')
 
 exports.handler = async (event) => {
+  // Reason for this check: event.body is a string of payload when AWS API Gateway is used.
```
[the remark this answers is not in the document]
butenkor: Maybe better: `// Reason for this check: if AWS API Gateway is used then event.body is provided as a string payload.`

notification/src/handler/notification/notification.handler.js
```diff
@@ -112,6 +132,21 @@ async function updatePaymentWithRepeater(
   }
 }
 
+function _obfuscateNotificationInfoFromActionFields(updateActions) {
```
butenkor, line 135: Why do we need that? What do we attempt to filter out?

notification/src/handler/notification/notification.handler.js
```diff
@@ -83,23 +94,32 @@ async function updatePaymentWithRepeater(
...
-      if (err.body.statusCode !== 409)
-        throw new Error(
-          `Unexpected error during updating a payment with ID: ${currentPayment.id}. Exiting. ` +
-            `Error: ${JSON.stringify(serializeError(err))}`
-        )
+      if (err.statusCode !== 409)
+        throw new CommercetoolsError({
+          stack: err.stack,
+          message:
+            `Unexpected error during updating a payment with ID: ${currentPayment.id}. Exiting. ` +
```
butenkor, line 101:
```suggestion
            `Unexpected error on payment update with ID: ${currentPayment.id}. ` +
```

notification/src/handler/notification/notification.handler.js
```diff
@@ -83,23 +94,32 @@ async function updatePaymentWithRepeater(
...
+            `Actions: ${JSON.stringify(
```
butenkor, line 102:
```suggestion
            `Failed actions: ${JSON.stringify(
```

notification/src/handler/notification/notification.handler.js
ahmetoz: `addInterfaceInteraction` contains a whole notification object which could sensitive data in it, before we decided to not log all details of the notification payload. Here I figured out this part is not fixed, so added it as a boy scout kind of change.

notification/index.googleFunction.js
ahmetoz: Could you please check again: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/46a544d55af0abd27920873f58617b3f06a61217

notification/index.lambda.js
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/eb77efc032e6221e7fa9b4a6de8c80d83268e014

notification/src/handler/notification/notification.handler.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/6a553e7af08c1e8c86aaee23ac64ced8484ad171

notification/src/handler/notification/notification.handler.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/6a553e7af08c1e8c86aaee23ac64ced8484ad171

notification/src/handler/notification/notification.handler.js
```diff
@@ -112,6 +132,21 @@ async function updatePaymentWithRepeater(
   }
 }
 
+function _obfuscateNotificationInfoFromActionFields(updateActions) {
+  if (!updateActions) return updateActions
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/dab1c4f03e897ffd6fbc8224a9db17ff5ee4e152

JudeNiroshan (approved):

notification/src/handler/notification/notification.handler.js
butenkor: So what useful data it could potentially log if we remove all notification details?

notification/src/handler/notification/notification.handler.js
butenkor: Thanks. There is still a confusing/scary part saying "Exiting" :) Do we need it or what was the intention?

notification/src/handler/notification/notification.handler.js
```diff
@@ -6,6 +6,25 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
 
+class CommercetoolsError extends Error {
+  constructor({ stack, message, statusCode }) {
+    super()
+    this.stack = stack
+    this.message = message
+
+    /*
+      We do not want to block notifications coming by Adyen.
+      So with this error it will be accepted by notification module by default.
```
butenkor, line 17: confusing statement. Is default always the same or actually differ depending on isRecoverable state?

notification/src/handler/notification/notification.handler.js
```diff
@@ -6,6 +6,25 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
 const mainLogger = require('../../utils/logger').getLogger()
 
+class CommercetoolsError extends Error {
+  constructor({ stack, message, statusCode }) {
+    super()
+    this.stack = stack
+    this.message = message
+
+    /*
+      We do not want to block notifications coming by Adyen.
```
butenkor, line 16: We do not want but we do that if necessary. Why not just include the info:
```
recoverable: notification delivery can be retried by Adyen
non recoverable: notification delivery can not be retried by Adyen as it most probably would fail again

If commercetools status code is defined and is 5xx then return `500` to Adyen -> recoverable
If during communication with commercetools we got a `NetworkError` then return `500` -> recoverable
If commercetools status code is not OK but also not 5xx or 409 then return `accepted` -> non recoverable
```

notification/src/handler/notification/notification.handler.js
```diff
@@ -6,6 +6,25 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
+      So with this error it will be accepted by notification module by default.
+
+      Only we do not accept for statusCode === 0 (NetworkError) or 5xx errors like 500.
+      Note that, our client configured to do retries for statusCode=503 cases.
+      So when isRecoverable assigned to true the notification will not respond with accepted but with 500 status code.
+    */
+    this.isRecoverable =
+      statusCode !== undefined ? statusCode === 0 && statusCode >= 500 : false
```
butenkor, line 24: > statusCode === 0

Why statusCode can be 0? It is either not set due to network error or is defined, right?

Maybe something like that:

`this.isRecoverable = (statusCode !== undefined && statusCode > 399 && statusCode !== 409 && statusCode < 500 ) ? false : true`

notification/src/handler/notification/notification.handler.js
ahmetoz: One example might be, check the message, also in controller we put the notification into error context as it might be a different issue.

````
{
  "name": "ctp-adyen-integration-notifications",
  "hostname": "ct-00891.local",
  "pid": 90276,
  "level": 50,
  "notification": [
    {
      "eventCode": "AUTHORISATION",
      "eventDate": "2019-01-30T18:16:22+01:00",
      "pspReference": "test_AUTHORISATION_1",
      "success": "true"
    }
  ],
  "err": {
    "message": "Got a concurrent modification error when updating payment with id \"undefined\". Version tried \"undefined\", currentVersion: \"2\". Won't retry again because of a reached limit 20 max retries. Failed actions: [{\"action\":\"addInterfaceInteraction\",\"type\":{\"key\":\"ctp-adyen-integration-interaction-notification\",\"typeId\":\"type\"},\"fields\":{\"createdAt\":\"2021-03-25T06:40:33.161Z\",\"status\":\"authorisation\",\"type\":\"notification\",\"notification\":{\"eventCode\":\"AUTHORISATION\",\"eventDate\":\"2019-01-30T18:16:22+01:00\",\"pspReference\":\"test_AUTHORISATION_1\",\"success\":\"true\"}}},{\"action\":\"changeTransactionState\",\"transactionId\":\"9ca92d05-ba63-47dc-8f83-95b08d539646\",\"state\":\"Success\"}]",
    "isRecoverable": true
  },
  "msg": "Unexpected exception occurred.",
  "time": "2021-03-25T06:41:07.653Z",
  "v": 0
}


````

File: notification/src/handler/notification/notification.handler.js (hunk 9 above)
Comment (ahmetoz) [reply notification/src/handler/notification/notification.handler.js]: statusCode is sth set on the SDK, but you are right, I updated the statement (also 409 is added as recoverable)
https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/4cb04af997f9e9d4b57c48e8b310bac303f97ba4

File: notification/src/handler/notification/notification.handler.js (hunk 8 above)
Comment (ahmetoz) [reply notification/src/handler/notification/notification.handler.js]: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/67b2baa4eee6587bf075d3729cf0dc03a94e81cf

File: notification/src/handler/notification/notification.handler.js (hunk 7 above)
Comment (ahmetoz) [reply notification/src/handler/notification/notification.handler.js]: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/6258cb50e64867b16529d185444426d37d649ac8

File: notification/src/handler/notification/notification.handler.js [hunk 10]
```diff
@@ -6,6 +6,17 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
 const mainLogger = require('../../utils/logger').getLogger()
 
+class CommercetoolsError extends Error {
+  constructor({ stack, message, statusCode }) {
+    super()
+    this.stack = stack
+    this.message = message
+    // statusCode === 0 is NetworkError, default is recoverable.
+    this.isRecoverable =
+      statusCode !== undefined ? statusCode !== 0 && statusCode < 500 : true
```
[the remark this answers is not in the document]
ahmetoz: The last check is here: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/4cb04af997f9e9d4b57c48e8b310bac303f97ba4

notification/src/handler/notification/notification.handler.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/808a041ded5fabf89b52a706298200067d7773fc

notification/src/handler/notification/notification.handler.js
```diff
@@ -6,6 +6,31 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
+  constructor({ stack, message, statusCode }) {
+    super()
+    this.stack = stack
+    this.message = message
+    this.isRecoverable = this._isRecoverable(statusCode)
+  }
+
+  /**
+   * recoverable: notification delivery can be retried by Adyen
+   * non recoverable: notification delivery can not be retried by Adyen as it most probably would fail again
```
butenkor, line 19: I would add a line break here.

notification/src/handler/notification/notification.handler.js
```diff
@@ -6,6 +6,31 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
+   * If commercetools status code is defined and is 5xx then return `500` to Adyen -> recoverable
+   * If during communication with commercetools we got a `NetworkError` then return `500` -> recoverable
+   * If commercetools status code is not OK but also not 5xx or 409 then return `accepted` -> non recoverable
+   */
+  _isRecoverable(statusCode) {
```
butenkor, line 24: `isRecoverable` seems confusing to understand. Maybe `shouldRetry` or just `retry` would be better?

Sorry, I did not came with that earlier but some things just need multiple iterations :)

notification/src/handler/notification/notification.handler.js
```diff
@@ -6,6 +6,31 @@ const { getNotificationForTracking } = require('../../utils/commons')
...
+class CommercetoolsError extends Error {
+  constructor({ stack, message, statusCode }) {
+    super()
+    this.stack = stack
+    this.message = message
+    this.isRecoverable = this._isRecoverable(statusCode)
+  }
+
+  /**
+   * recoverable: notification delivery can be retried by Adyen
```
butenkor, line 18:
```suggestion
   * recoverable: notification delivery can be retried by Adyen (return 500)
```

notification/src/handler/notification/notification.handler.js
butenkor, line 19:
```suggestion
   * non recoverable: notification delivery can not be retried by Adyen as it most probably would fail again (return "accepted")
```

notification/src/handler/notification/notification.handler.js
```diff
@@ -83,23 +108,33 @@ async function updatePaymentWithRepeater(
...
       break
     } catch (err) {
-      if (err.body.statusCode !== 409)
-        throw new Error(
-          `Unexpected error during updating a payment with ID: ${currentPayment.id}. Exiting. ` +
-            `Error: ${JSON.stringify(serializeError(err))}`
-        )
+      if (err.statusCode !== 409)
+        throw new CommercetoolsError({
+          stack: err.stack,
```
butenkor, line 113: I wonder why we try to rewrite the error structure and not just include the error as object with all its properties?

notification/src/handler/notification/notification.handler.js
```diff
@@ -83,23 +108,33 @@ async function updatePaymentWithRepeater(
...
           'Got a concurrent modification error' +
           ` when updating payment with id "${currentPayment.id}".` +
           ` Version tried "${currentVersion}",` +
           ` currentVersion: "${err.body.errors[0].currentVersion}".`
-        throw new Error(
-          `${retryMessage} Won't retry again` +
+        throw new CommercetoolsError({
+          stack: err.stack,
+          message:
+            `${retryMessage} Won't retry again` +
```
butenkor, line 131: I noticed above that on retry we do recalculate update actions but we do not really check if newly update actions would be empty or not?

notification/src/handler/notification/notification.handler.js
butenkor: I see thanks. I wonder is it just the test or the payment id might be always "undefined"?

lojzatran (approved):

notification/src/handler/notification/notification.handler.js
ahmetoz: No it's by chance, mock does not have both version and id, mock is a draft :)

notification/src/handler/notification/notification.handler.js
ahmetoz: 5e48180

notification/src/handler/notification/notification.handler.js
ahmetoz: we could and capture it, but it has original requests and lots of details, I guess the reason was the sensitive data but it's also useful for tracking. Not sure the way to go here.

notification/src/handler/notification/notification.handler.js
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/fddb3184ada4c007e2b24d4b5753e385a008b90c

notification/src/handler/notification/notification.handler.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/eb7cd73f4ed646a488e9c7f115292f7a561a3e0d

notification/src/handler/notification/notification.handler.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/698ae1e0492ff9c74e1cb6cde34b7d16a29647ef

ahmetoz: looks like a bug, good catch: https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/0d3ed73a826676ff40857781c518026c3ffd74d6

ahmetoz:
> Remove the negation:
> 
> ```
> statusCode === undefined || statusCode < 400 || statusCode === 409 || statusCode >= 500
> ```

https://github.com/commercetools/commercetools-adyen-integration/pull/629/commits/fddb3184ada4c007e2b24d4b5753e385a008b90c

notification/src/handler/notification/notification.handler.js
ahmetoz: no worries, I am trying to keep changes to a minimum as PR is not focusing on the refactoring much, thanks for the suggestions.

notification/src/handler/notification/notification.handler.js
butenkor: Why not just:

```
throw new CommercetoolsError({
          error: err,
```
?

butenkor: btw. this library seem promising as it allows to re-throw with original error details: https://www.npmjs.com/package/verror

ahmetoz: The request is added to the backlog and will be added soon.

PR #630, opened 2021-03-23
Remove stacktrace in error response returned from GCP cloud function

leungkinghin:
- Remove stacktrace in error message returned from GCP cloud function.
- Return error response quickly when request is coming from browser, which brings an invalid payload.


extension/index.googleFunction.js
```diff
@@ -8,6 +8,17 @@ const logger = utils.getLogger()
 exports.extensionTrigger = async (request, response) => {
   try {
     const authToken = auth.getAuthorizationRequestHeader(request)
+    if (!request?.body?.resource) {
```
ahmetoz, line 11: What returns when `auth.getAuthorizationRequestHeader` throws an error ? Any test case ?

ahmetoz (commented):
Is the scope only for google function or in general ?

extension/index.googleFunction.js
```diff
@@ -20,12 +31,10 @@ exports.extensionTrigger = async (request, response) => {
       })
     }
   } catch (err) {
-    const errorMessage = `Unexpected error: ${JSON.stringify(
-      serializeError(err)
```
lojzatran, line 24: Remove the require part for serializeError in the first line.

extension/index.googleFunction.js
lojzatran, line 11: const obj = request?.body?.resource?.obj

extension/index.googleFunction.js
```diff
@@ -8,6 +8,17 @@ const logger = utils.getLogger()
 exports.extensionTrigger = async (request, response) => {
   try {
     const authToken = auth.getAuthorizationRequestHeader(request)
+    if (!request?.body?.resource) {
+      response.status(400).send({
```
lojzatran, line 12: Use return everywhere and don't omit it. Otherwise you could have weird behaviour like here where you don't use return - it will send 400, but the code will continue.

lojzatran (changes requested):

extension/index.googleFunction.js
```diff
@@ -20,12 +31,10 @@ exports.extensionTrigger = async (request, response) => {
...
     }
   } catch (err) {
-    const errorMessage = `Unexpected error: ${JSON.stringify(
-      serializeError(err)
-    )}`
+    const errorMessage = `Unexpected error: ${err.message}`
     logger.error(errorMessage)
 
-    response.status(500).send({
+    response.status(400).send({
```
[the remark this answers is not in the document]
leungkinghin: CTP only accept 200 or 400 https status from extension.
https://docs.commercetools.com/api/projects/api-extensions#response

If extension returns 500 status code from extension to CTP, CTP returns "500 is not an expected status code." in error message.

extension/index.googleFunction.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/ba3a6b0ca39bab097f093e6a96d06635d135e8a0

extension/index.googleFunction.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/67c6a4aba8f26742fa6793ab4d82b543fa8e470e

extension/index.googleFunction.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/8b9884d0f3fd366b1ec5feb1d318f7a05432ecd5

extension/index.googleFunction.js
leungkinghin: It is required now for logging purpose

extension/index.googleFunction.js
```diff
@@ -20,12 +31,10 @@ exports.extensionTrigger = async (request, response) => {
       })
     }
   } catch (err) {
-    const errorMessage = `Unexpected error: ${JSON.stringify(
-      serializeError(err)
-    )}`
+    const errorMessage = `Unexpected error: ${err.message}`
```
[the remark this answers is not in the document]
leungkinghin: Log the stacktrace now but not returning to user.
https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/63677d02f13bb21af8a52db9759785a07851666a

extension/index.googleFunction.js
leungkinghin: it should be caught here
https://github.com/commercetools/commercetools-adyen-integration/blob/ffc9c0db114fc33e3d730832fff0539acc04ca9e/extension/index.googleFunction.js#L31

so far we don't have unit test for google cloud HTTP function. It may need another unit test framework
https://www.deloittedigital.at/google-cloud-functions-testing-part-1/

leungkinghin:
> Is the scope only for google function or in general ?

For GCP only. Let me edit the title to avoid confusion.

lojzatran (approved):

extension/index.googleFunction.js
leungkinghin: Added a unit test for GCP extension trigger
https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/3540c47391dd9c16aa8f74df602a268d7f4df92d

extension/index.googleFunction.js
```diff
@@ -7,31 +7,57 @@ const logger = utils.getLogger()
...
+      response,
+      statusCode: 400,
+      body: {
+        errors: paymentResult.data ? paymentResult.data.errors : undefined,
+      },
+    })
   } catch (err) {
-    const errorMessage = `Unexpected error: ${JSON.stringify(
+    const errorMessage = `Unexpected error: ${err.message}`
+    const errorStackTrace = `Unexpected error: ${JSON.stringify(
```
butenkor, line 45: I would always try to include payment ID from the `obj` in the logged stack traces.

extension/index.googleFunction.js
```diff
@@ -7,31 +7,57 @@ const logger = utils.getLogger()
...
+        body: {
+          actions: paymentResult.data ? paymentResult.data.actions : [],
+        },
       })
     }
+    return utils.sendGoogleFunctionResponse({
+      response,
+      statusCode: 400,
+      body: {
+        errors: paymentResult.data ? paymentResult.data.errors : undefined,
```
butenkor, line 40: I am curious why `errors` are nested within data? Why existence of data implies those are errors? Looks that there might be also update actions instead of errors: https://github.com/commercetools/commercetools-adyen-integration/blob/3d752e6dbf5a62febe288082b801a256ff07d3fa/extension/src/paymentHandler/payment-handler.js#L58

 Would be simpler if response object would look like that:
```
{
  "actions": [] // defined in case there are any
  "errors": [] // defined in case there are any
}
```

extension/src/utils.js
```diff
@@ -23,6 +23,10 @@ function sendResponse({ response, statusCode = 200, headers, data }) {
   response.end(JSON.stringify(data))
 }
 
+function sendGoogleFunctionResponse({ response, statusCode = 200, body }) {
+  response.status(statusCode).send(body)
+}
```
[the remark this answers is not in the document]
leungkinghin: it makes easier to create a stub in unit test case, same approach as following
https://github.com/commercetools/commercetools-adyen-integration/blob/3540c47391dd9c16aa8f74df602a268d7f4df92d/extension/test/unit/payment.controller.spec.js#L11

extension/index.googleFunction.js
```diff
@@ -7,31 +7,57 @@ const logger = utils.getLogger()
...
-        },
-      ],
+    return utils.sendGoogleFunctionResponse({
+      response,
+      statusCode: 400,
+      body: {
+        errors: [
+          {
+            code: 'InvalidOperation',
+            message: errorMessage,
```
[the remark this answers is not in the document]
leungkinghin: Yes. we can change to `BadRequest` here
https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/1895bf5cfd68993a2ca29daa9330fa23cf6da7e7

extension/index.googleFunction.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/60231a2c7d727dace55bde76e966374a30fb59e1

extension/index.googleFunction.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/2149fca06f83665f83c7bb39a36feedc99bb10d9

leungkinghin: @butenkor just realized that the above change would fail e2e test all the time. Shall we bypass it at this moment?

extension/index.googleFunction.js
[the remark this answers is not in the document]
leungkinghin: CTP doesn't accept `BadRequest` as error code, otherwise CTP throws exception to user.
In this exception, the stack-trace shows a list of error code it can support. `InvalidOperation` is one of them.

extension/src/utils.js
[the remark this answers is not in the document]
leungkinghin: Removed the utility function and made a mock resposne in unit test 
https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/3f09be69d2d2be3ae16dcc6ff44f5cc82a6b9fe9

extension/index.googleFunction.js
butenkor: We have it covered in the follow-up ticket.

extension/index.googleFunction.js
```diff
@@ -6,29 +6,39 @@ const auth = require('./src/validator/authentication')
...
     }
+    return response.status(400).send({
+      errors: paymentResult.data ? paymentResult.data.errors : undefined,
+    })
   } catch (err) {
-    const errorMessage = `Unexpected error: ${JSON.stringify(
-      serializeError(err)
-    )}`
-    logger.error(errorMessage)
+    const errorMessage = `Unexpected error (Payment ID: ${obj?.id}): ${err.message}. `
```
ahmetoz, line 32:
```suggestion
    const errorMessage = `Unexpected error (Payment ID: ${obj?.id}): ${err.message}.`
```

extension/index.googleFunction.js
```diff
@@ -6,29 +6,39 @@ const auth = require('./src/validator/authentication')
 const logger = utils.getLogger()
 
 exports.extensionTrigger = async (request, response) => {
+  const obj = request?.body?.resource?.obj
```
ahmetoz, line 9: Would be nice if those consistent between handlers?

```suggestion
  const paymentObj = request?.body?.resource?.obj
```

extension/index.lambda.js
```diff
@@ -22,6 +35,15 @@ exports.handler = async (event) => {
     }
   } catch (e) {
     logger.error(e, `Unexpected error when processing event`)
-    throw e
+    const errorObj = {
```
ahmetoz, line 38: Would be nice to use the same pattern for an error message like you're doing for index.googleFunction, as those error handlers duplicate in a different part of the module (payment.controller and index.lambda), would be nice to find a place same parts of it a function, you might return the result, for index.lambda you might merge/extend `responseType: 'FailedValidation',` inside the request. What do you think?

````js
} catch (err) {
    const errorMessage = `Unexpected error (Payment ID: ${paymentObj?.id}): ${err.message}. `
    const errorStackTrace = `Unexpected error (Payment ID: ${
      paymentObj?.id
    }): ${JSON.stringify(serializeError(err))}`
    logger.error(errorStackTrace)

    return {
      responseType: 'FailedValidation',
      errors: [
        {
          code: 'General',
          message: errorMessage,
        },
      ],
    }
  }
````

**Review by ahmetoz — CHANGES_REQUESTED** [review]

File: extension/index.googleFunction.js (hunk 11 above)
Comment (leungkinghin) [reply extension/index.googleFunction.js]: https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/83cbc9ac399631e7308e35593a0b33d3d3036d48

File: extension/index.lambda.js (hunk 12 above)
Comment (leungkinghin) [reply extension/index.lambda.js]: Yup. You're right. It can be refactored.
https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/b219a065378b95602f19a09bfd9beee3c59a6ec7

**Comment by leungkinghin** [comment]
> could you remove debugger logs ?

https://github.com/commercetools/commercetools-adyen-integration/pull/630/commits/b47e417f592da4b98650f1b144a5b13f94bf06f5

**Review by ahmetoz — APPROVED** [review]

File: docs/ContributionGuide.md [hunk 13]
```diff
@@ -20,7 +20,7 @@
 ## Prerequisites
 
 Minimum requirements are:
- - **Node.js** version 12.
+ - **Node.js** version 14.
```
butenkor, line 23: Is 14 really our minimum? I find it always better if it says for example 12 and above for example.

butenkor (approved):

docs/ContributionGuide.md
leungkinghin: Yes. we have written Optional Chaining in the code, which is supported in node 14 or above.

PR #659, opened 2021-04-21
Adyen v67

### LINKED ISSUE #658: Support for Adyen v67 API
Hello,
I've used to work on integration with v52 API of Adyen. However, I see now, we should use the newest version of both Web Drop-In and Adyen's API. That's why I am trying to apply support for v67 (latest) version of API.

The first difference I found is that I cannot pass any additional (not specified in doc) property anymore because Adyen throws a validation error. The thing is you extend `submitAdditionalPaymentDetails` with `applicationInfo` which is redundant field - not possible to pass validation anymore in the newest version of API:
```ts
function submitAdditionalPaymentDetails(
  merchantAccount,
  commercetoolsProjectKey,
  submitAdditionalPaymentDetailsRequestObj
) {
  const adyenCredentials = config.getAdyenConfig(merchantAccount)
  extendRequestObjWithMetadata(
    submitAdditionalPaymentDetailsRequestObj,
    commercetoolsProjectKey
  )
  extendRequestObjWithApplicationInfo(submitAdditionalPaymentDetailsRequestObj)
  return callAdyen(
    `${adyenCredentials.apiBaseUrl}/payments/details`,
    merchantAccount,
    adyenCredentials.apiKey,
    submitAdditionalPaymentDetailsRequestObj
  )
}
```

image

I am going to play with it and prepare a PR if succeed.

Fifciu:
closes #658 

Indeed, it looks like `submitAdditionalPaymentDetails` does not work with v67 API because we add `additionalInfo` field which is not accepted by Adyen's validator. Problem appears for each payment method that uses this request (3ds cards, paypal - tested manually by me).

This PR fixes the problem.


lojzatran:
Hi @Fifciu,

Thank you for the PR. You're right about the issue and the fix looks good. I will try to merge it to our branch instead of master and run the tests. If they pass, I will merge to master and make a release.

PR #665, opened 2021-04-23
Shorten metadata.commercetoolsProjectKey name

lojzatran:
Due to restrictions in the Adyen API, it's not allowed to have metadata key with length more than 20 characters and metadata values with more than 80 characters. In this PR, we fixed this problem.

### DIFF
--- a/extension/src/service/web-component-service.js
+++ b/extension/src/service/web-component-service.js
@@ -104,6 +104,8 @@ function extendRequestObjWithApplicationInfo(requestObj) {
 
 function extendRequestObjWithMetadata(requestObj, commercetoolsProjectKey) {
   requestObj.metadata = {
+    // metadata key must have length of max. 20 chars
+    // metadata value must have length of max. 80 chars
     ctProjectKey: commercetoolsProjectKey,
   }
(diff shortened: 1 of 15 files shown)


extension/src/service/web-component-service.js
```diff
@@ -104,7 +104,7 @@ function extendRequestObjWithApplicationInfo(requestObj) {
 
 function extendRequestObjWithMetadata(requestObj, commercetoolsProjectKey) {
   requestObj.metadata = {
-    commercetoolsProjectKey,
+    ctProjectKey: commercetoolsProjectKey,
```
butenkor, line 107: Maybe we should document here the restrictions on the length of the metadata fields?

lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/665/commits/f68c497458000839c6e212c4d75a15eefb749a00

butenkor (approved):

Fifciu:
I've just applied this fix for my integrations. Work like a charm, great job @lojzatran !

PR #673, opened 2021-04-28
Use newest adyen api

lojzatran:
This PR contains a breaking change so I would increase the major version.

# What has changed
- For the extension module, the only real change is [the URL in the config.js](https://github.com/commercetools/commercetools-adyen-integration/pull/673/files#diff-f25ee3db61b811f4517cf9bfb92ef8d3caff17298d34a379448a310a4921beea)

### The rest are only test changes:

- This change means the redirect is not done by POST anymore, but with GET. Additionally, the params are not passed anymore as POST body, but in a GET request params.
   - https://docs.adyen.com/online-payments/release-notes?tab=script_3ad58c0d_2#breaking-changes
For redirect 3D Secure authentication, you now need to submit the redirectResult appended to the returnUrl when the shopper is redirected back to you. This replaces the MD and the PaRes parameters used in previous versions, and aligns the flow for 3D Secure redirects with all other redirects.
For redirect 3D Secure authentication, the shopper is now redirected back to you from the issuer with an HTTP GET instead of HTTP POST. The returnURL is appended with a redirectResult.
 (https://github.com/commercetools/commercetools-adyen-integration/pull/673/files#diff-c11f3b6b00673590b4fa686a12ac39e6ca28602583e77e80ad81c0dcde2710bf)

- The JSON structre from `getPaymentMethods` is changed. I also updated Adyen JS library to be sure it will work with the new JSON structure.
   - https://docs.adyen.com/online-payments/release-notes#removed-deprecated-fields-1
`groups` attr for `/paymentMethods` was removed. (https://github.com/commercetools/commercetools-adyen-integration/pull/673/files#diff-2b796db286f8860b9408209144668a7e27d336329b53dc1df688f59999303167)

- This change means one step is removed for 3DS native. There's no need to do 2x `submitAdditionalPaymentDetails` anymore.
  - https://docs.adyen.com/online-payments/release-notes?tab=script_3ad58c0d_2#breaking-changes
For native 3D Secure 2 authentication, the response can no longer contain an action object. This means that native 3D Secure 2 payments no longer require multiple /payments/details requests. (https://github.com/commercetools/commercetools-adyen-integration/pull/673/files#diff-c3e0e34a31ba18387891bf4c47a15950e3741501a4d6bb82bb44e7bb089d1b3f)


extension/test/unit/config/config.spec.js
```diff
@@ -83,10 +83,10 @@ describe('::config::', () => {
       projectKey: 'ctpProjectKey1',
     })
     expect(config.getAdyenConfig('adyenMerchantAccount1')).to.eql({
-      apiBaseUrl: 'https://checkout-test.adyen.com/v52',
+      apiBaseUrl: 'https://checkout-test.adyen.com/v67',
       apiKey: 'apiKey',
       clientKey: 'clientKey',
-      legacyApiBaseUrl: 'https://pal-test.adyen.com/pal/servlet/Payment/v52',
+      legacyApiBaseUrl: 'https://pal-test.adyen.com/pal/servlet/Payment/v64',
```
butenkor, line 89: Why not v67?

extension/src/config/config.js
```diff
@@ -66,10 +66,10 @@ function getAdyenConfig(adyenMerchantAccount) {
     )
   return {
     apiKey: adyenConfig.apiKey,
-    apiBaseUrl: adyenConfig.apiBaseUrl || 'https://checkout-test.adyen.com/v52',
+    apiBaseUrl: adyenConfig.apiBaseUrl || 'https://checkout-test.adyen.com/v67',
     legacyApiBaseUrl:
       adyenConfig.legacyApiBaseUrl ||
-      'https://pal-test.adyen.com/pal/servlet/Payment/v52',
+      'https://pal-test.adyen.com/pal/servlet/Payment/v64',
```
butenkor, line 72: v64?

butenkor (commented):
Thanks. Please make sure that we mention breaking changes (if there are any) and document some migration hints on release.

extension/test/unit/config/config.spec.js
lojzatran: This is different endpoint and here they have only v64: https://docs.adyen.com/api-explorer/#/Payment/v64/overview

extension/src/config/config.js
lojzatran: This is different endpoint and here they have only v64: https://docs.adyen.com/api-explorer/#/Payment/v64/overview

butenkor (approved):

butenkor:
> This change means one step is removed for 3DS native. There's no need to do 2x submitAdditionalPaymentDetails anymore.

Do we need to adapt the doc [here](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/WebComponentsIntegrationGuide.md#action-response-1) or this second submitAdditionalPaymentDetails is not not required for 3ds process?

lojzatran:
> > This change means one step is removed for 3DS native. There's no need to do 2x submitAdditionalPaymentDetails anymore.
> 
> Do we need to adapt the doc [here](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/WebComponentsIntegrationGuide.md#action-response-1) or this second submitAdditionalPaymentDetails is only not required for 3ds process?

This is general step, not only 3ds specific.

Issue #700, opened 2021-05-20
Support custom shipping method use with Klarna payment


pawelphilipczyk:
I've got extension up and running and I'm using Adyen dropin component. Card payments are working fine. 

Now the problem starts with Klarna payments and specifically in combination with custom shipping methods in commerce tools. I've managed to track down the issue based on the error below to `_getShippingMethodDescription` function in `Klarna-make-payment.handler.js`. Seems that it's trying to read shipping method description based on reference object and that is probably working only for regular shipping methods not the custom ones. 

This bug is probably easy to fix but I'm wondering why is it doing that anyway? 
Examples in Adyen documentation for Klarna do not seem to have any of this data in the payloads. 

`"Unexpected error (Payment ID: 92164059-4974-4015-a613-5da27de301e3): {\"name\":\"TypeError\",\"message\":\"Cannot read property 'obj' of undefined\",\"stack\":\"TypeError: Cannot read property 'obj' of undefined\\n    at _getShippingMethodDescription (/var/task/src/paymentHandler/klarna-make-payment.handler.js:128:54)\\n    at _createShippingInfoAdyenLineItem (/var/task/src/paymentHandler/klarna-make-payment.handler.js:121:18)\\n    at createLineItems (/var/task/src/paymentHandler/klarna-make-payment.handler.js:58:20)\\n    at Object.execute (/var/task/src/paymentHandler/klarna-make-payment.handler.js:21:41)\\n    at processTicksAndRejections (internal/process/task_queues.js:93:5)\\n    at async Promise.all (index 0)\\n    at async Object.handlePayment (/var/task/src/paymentHandler/payment-handler.js:51:28)\\n    at async Runtime.exports.handler (/var/task/index.js:24:27)\"}",
    "time": "2021-05-20T08:08:42.578Z",`

lojzatran:
Hi @pawelphilipczyk,

The reason shipping method is also included within line items is that from our previous experience with Klarna you have to include shipping costs in the payment too. If not, the final price that the customer pays will be lower. We use this link to add shipping to the payment: https://docs.adyen.com/payment-methods/klarna/web-component#make-a-payment-request

lojzatran:
You're right about the bug. We will check within the team when it can be fixed, but you're welcome to provide a PR for that! 😉

pawelphilipczyk:
[...]

PR #701, opened 2021-05-20
Fix klarna custom shipping method description getter

pawelphilipczyk:
As described in #700 currently Klarna payment fails for custom shipping methods.

I can update this PR if decision is made to use `name` or `localizedDescription` for standard shipping methods instead of deprecated `description`. See docs related to this: https://docs.commercetools.com/api/projects/shippingMethods#shippingmethod

### DIFF
--- a/extension/src/paymentHandler/klarna-make-payment.handler.js
+++ b/extension/src/paymentHandler/klarna-make-payment.handler.js
@@ -125,7 +125,7 @@ function _createShippingInfoAdyenLineItem(shippingInfo) {
 }
 
 function _getShippingMethodDescription(shippingInfo) {
-  const shippingMethod = shippingInfo.shippingMethod.obj
+  const shippingMethod = shippingInfo.shippingMethod?.obj
   if (shippingMethod) return shippingMethod.description
   return shippingInfo.shippingMethodName
 }
--- a/extension/src/paymentHandler/klarna-make-payment.handler.js
+++ b/extension/src/paymentHandler/klarna-make-payment.handler.js
@@ -5,6 +5,7 @@ const config = require('../config/config')
 const ADYEN_PERCENTAGE_MINOR_UNIT = 10000
 const DEFAULT_PAYMENT_LANGUAGE = 'en'
 const KLARNA_DEFAULT_LINE_ITEM_NAME = 'item'
+CONST KLARNA_DEFAULT_SHIPPING_METHOD_DESCRIPTION = 'shipping'
 
 async function execute(paymentObject) {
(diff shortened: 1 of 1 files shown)


lojzatran:
Thanks for the PR! Your suggestion sounds good. I have a suggestion for you, WDYT?

First check for `localizedDescription` field - choose the correct locale using the locale from klarna method (https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/src/paymentHandler/klarna-make-payment.handler.js#L68). If this field doesn’t exist, then fall back to `description`. And if that doesn’t `description`, then fall back to `name` (this part is already done).

In case of custom method use `shippingMethodName` as you suggested.

pawelphilipczyk:
@lojzatran I've updated it but I'm not sure it's what you meant. The level of "fallbacking" here gets complicated ;) I'm trying to use `_localizeOrFallback` here with fallback to deprecated `description`. 

It will return some falsy value if both are empty so I've added additional fallback with `KLARNA_DEFAULT_SHIPPING_METHOD_DESCRIPTION` outside of that function. 

That will also handle the case if we have custom shipping method and shippingMethodName is not available for some reason? I'm not sure if that can ever happen because that property is required in `setCustomShippingMethod` action but maybe it can be empty string. 

I haven't tested it yet. Do you have some tests for it or quick way to test "manually"? I will do some testing next week but we don't have all those cases.

lojzatran:
@pawelphilipczyk sorry for the late reply. We don't have a quick way to manually test it, but we have automated tests. Unfortunately they are not triggered for the forks. It would be good if you can also add unit/integration tests for this change. If it's too much, I can also take over the PR and try finish it.

pawelphilipczyk:
@lojzatran I can give it a try. 

Just to point me in the right direction, I'm assuming this is where it should be added: `commercetools-adyen-integration/extension/test/unit/klarna-make-payment.handler.spec.js`?

lojzatran:
> @lojzatran I can give it a try.
> 
> Just to point me in the right direction, I'm assuming this is where it should be added: `commercetools-adyen-integration/extension/test/unit/klarna-make-payment.handler.spec.js`?

Yes, I would try to add it there like a new unit test.

lojzatran:
Hi @pawelphilipczyk 
Do you have any news here? If not, I will start working on it and try to finish it within this week.
Thank you.

pawelphilipczyk:
Unfortunately I have had to switch to other more important tasks before I get back to this (which will be not until few weeks from now on). 
I'll be happy to test whatever fix you provide.

PR #732, opened 2021-06-14
Add debug logs to notification module

lojzatran:


notification/src/api/notification/notification.controller.js
```diff
@@ -37,6 +43,18 @@ async function handleNotification(request, response) {
   }
 }
 
+function logNotificationForDebug(notification) {
+  const notificationCloned = _.cloneDeep(notification)
```
butenkor, line 47: Do we really need to obfuscate the data if debug mode is going to be used rather during the tests only and we attempt to see more info?

butenkor (commented):
Maybe it would be also useful to log as debug what extension module gets from the API and what actions it generates as response?

lojzatran:
> Maybe it would be also useful to log as debug what extension module gets from the API and what actions it generates as response?

https://github.com/commercetools/commercetools-adyen-integration/pull/732/commits/fc39dc0d065b59f7b86b9b827c408509b782532a

notification/src/api/notification/notification.controller.js
lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/732/commits/d3bb8c9e6082bad0edfdeb411be56644e2514b00

extension/src/api/payment/payment.controller.js
```diff
@@ -38,6 +43,8 @@ async function processRequest(request, response) {
         : { errors: paymentResult.errors },
     }
 
+    logger.debug('Returned data', JSON.stringify(result.data))
```
butenkor, line 46:
```suggestion
    logger.debug('Data to be returned', JSON.stringify(result.data))
```

butenkor (approved):
Looks good. Just added small wording change.

PR #733, opened 2021-06-15
Bugfix/klarna custom shipping method

lojzatran:
Continuation of an existing PR: https://github.com/commercetools/commercetools-adyen-integration/pull/701

Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/700

## How it should work:
first check for localizedDescription field, if it doesn’t exist, then fall back to description. And if that doesn’t exist, then fall back to name. In case of custom method use shippingMethodName.

## Extra fix: use the first available locale value instead of a default language.

### DIFF
--- a/extension/src/paymentHandler/klarna-make-payment.handler.js
+++ b/extension/src/paymentHandler/klarna-make-payment.handler.js
@@ -5,7 +5,7 @@ const config = require('../config/config')
 const ADYEN_PERCENTAGE_MINOR_UNIT = 10000
 const DEFAULT_PAYMENT_LANGUAGE = 'en'
 const KLARNA_DEFAULT_LINE_ITEM_NAME = 'item'
-CONST KLARNA_DEFAULT_SHIPPING_METHOD_DESCRIPTION = 'shipping'
+const KLARNA_DEFAULT_SHIPPING_METHOD_DESCRIPTION = 'shipping'
 
 async function execute(paymentObject) {
   const makePaymentRequestObj = JSON.parse(
--- a/extension/src/paymentHandler/klarna-make-payment.handler.js
+++ b/extension/src/paymentHandler/klarna-make-payment.handler.js
@@ -61,11 +61,6 @@ function createLineItems(payment, cart) {
   return lineItems
 }
 
-/**
- * There will always be a locale `DEFAULT_PAYMENT_LANGUAGE` as a default fallback.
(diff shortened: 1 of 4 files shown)


extension/src/paymentHandler/klarna-make-payment.handler.js
```diff
@@ -1,10 +1,11 @@
+const _ = require('lodash')
 const ctpClientBuilder = require('../ctp')
 const makePaymentHandler = require('./make-payment.handler')
 const config = require('../config/config')
 
 const ADYEN_PERCENTAGE_MINOR_UNIT = 10000
-const DEFAULT_PAYMENT_LANGUAGE = 'en'
```
ahmetoz, line 6: it seems that the doc of _getLocales function in this file is not up to date. see:
```...There will always be a locale `DEFAULT_PAYMENT_LANGUAGE` as a default fallback. ```
similar for ```If it's not set, it will pick `DEFAULT_PAYMENT_LANGUAGE`.```

File: extension/src/paymentHandler/klarna-make-payment.handler.js [hunk 2]
```diff
@@ -114,25 +112,37 @@ function _createAdyenLineItemFromCustomLineItem(ctpLineItem, locales) {
...
+      shippingMethod.description
+    )
+  }
   return shippingInfo.shippingMethodName
 }
 
 function _localizeOrFallback(localizedString, locales, fallback) {
-  const locale = locales.find((l) => localizedString[l])
-  return locale ? localizedString[locale] : fallback
+  let result
```
ahmetoz, line 140: I guess the additional `result` variable is not needed, maybe it could be more readable without it.

ahmetoz (commented):
Looks good.

extension/src/paymentHandler/klarna-make-payment.handler.js
lojzatran: Do you mean to return immediately? I don't think it makes a difference here, so I'll just keep it as it is.

extension/src/paymentHandler/klarna-make-payment.handler.js
lojzatran: I removed the comments as they're not very useful: https://github.com/commercetools/commercetools-adyen-integration/pull/733/commits/683941ed15303d093814c97e98d26ef91b7a91d6

ahmetoz (approved):
🎉

PR #745, opened 2021-06-29
Make removal of the sensitive data optional

lojzatran:
**Situation**
Currently we remove data from payloads if we account them as sensitive.

**Complication**
There are cases where important checkout informations are hidden within those sensitive informations and through automatic deletion by the integration are not available for the further decision making of the front end.

**Resolution**
Make it possible to deactivate the removal of the sensitive data but the removal should be by default activated.


extension/docs/HowToRun.md
```diff
@@ -147,6 +148,7 @@ Other configurations can be set as direct child attributes in `ADYEN_INTEGRATION
...
 | `keepAliveTimeout`           | Milliseconds to keep a socket alive after the last response ([Node.js docs](https://nodejs.org/dist/latest/docs/api/http.html#http_server_keepalivetimeout)).                      | NO       | Node.js default (5 seconds)                                                                           |
 | `basicAuth`                  | Boolean attribute to enable/disable basic authentication to prevent unauthorized 3rd-party from accessing extension endpoint                                                       | NO       | false                                                                                                 |
+| `logSensitiveData`           | Boolean attribute. When `true`, all data from Adyen will be saved in the interface interaction and in the custom fields                                                            | NO       | false                                                                                                 |
```
ahmetoz, line 151: `log` prefix might confusing for users, I guess logic is more related to removing additional data, so I would use `removeAdditionalData` and default value as true. 

For description, 

Boolean attribute. When set to `false`, [additionalData](https://docs.adyen.com/api-explorer/#/CheckoutService/v67/post/payments__resParam_additionalData) which contains additional information about the payment returning from the Adyen responses `will be saved` in the interface interaction and in the custom fields.

notification/src/handler/notification/notification.handler.js
```diff
@@ -253,13 +256,17 @@ function compareTransactionStates(currentState, newState) {
...
 function getAddInterfaceInteractionUpdateAction(notification) {
-  // strip away sensitive data
-  delete notification.additionalData
-  delete notification.reason
+  const moduleConfig = config.getModuleConfig()
+  const notificationToUse = _.cloneDeep(notification)
+  if (!moduleConfig.logSensitiveData) {
+    // strip away sensitive data
+    delete notificationToUse.NotificationRequestItem.additionalData
+    delete notificationToUse.NotificationRequestItem.reason
```
ahmetoz, line 264: here we are removing `reason` but do not ensuring it in test cases, also I'm not really sure if the reason is sensitive data, it might something related to the refusals. Would be better to ask Adyen imo. I see some things also in this doc, search with `reason` : https://docs.adyen.com/development-resources/webhooks/understand-notifications

ahmetoz (commented):
Overall looks good for the coding perspective, I have some comments for the functionality.

notification/src/handler/notification/notification.handler.js
lojzatran: This code is old, I guess there is a reason that `reason` is considered as sensitive data. I added verification for it: https://github.com/commercetools/commercetools-adyen-integration/pull/745/commits/d75b0510387a74d4890f4c44327b71666b38cca6

extension/docs/HowToRun.md
lojzatran: Sounds the same to me :) Anyway, we don't remove only additionalData and the idea of this field is to remove sensitive data, so I change the name to `removeSensitiveData`: https://github.com/commercetools/commercetools-adyen-integration/pull/745/commits/f0e09786af71fb0ac001e8f609f197014b20a4d9

ahmetoz (approved):
👍🏽

PR #786, opened 2021-08-06
cancelPayment does not transmit project name

jvoges:
the notifcation module can't update cancelled payments as the notification payload is missing the project name.
this fix adds it to the request payload in the extension module

### DIFF
--- a/extension/src/service/web-component-service.js
+++ b/extension/src/service/web-component-service.js
@@ -68,7 +68,7 @@ function cancelPayment(
   cancelPaymentRequestObj
 ) {
   const adyenCredentials = config.getAdyenConfig(merchantAccount)
-  extendRequestObjWithMetadata(cancelPaymentRequestObj)
+  extendRequestObjWithMetadata(cancelPaymentRequestObj, commercetoolsProjectKey)
   return callAdyen(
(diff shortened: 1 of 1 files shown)


lojzatran:
Thank you for the fix @jvoges. I tried your fix in adyen sandbox env, but unfortunately the notifications still come without the project name. Does this fix work for you?

I'm currently trying to check with Adyen why is it so and will let you know once we have more information.

peniakoff:
@lojzatran it looks like the Adyen doesn't send metadata with every notification call. I don't know yet, why...

butenkor:
We expect a fix from Adyen side to be rolled out in the upcoming weeks which should ensure that metadata is included on all the required notifications.

butenkor:
Hi @jvoges I am closing the PR since it does not fix the issue. Also, Adyen is still working on ensuring that metadata are always included on all the notifications but it will take time. Until it is ensured from Adyen side it is possible to work around the issue by incorporating commercetools project key in the Adyen notification/webhook URL and is possible since the release [v9.8.0](https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v9.8.0)

PR #799, opened 2021-08-12
Renovate config

praveenkumarct:
* Disable updates in the docker file. (Ex: avoid updating node versions)
* Batch them all into 1 PR weekly.
* StabilityDays:
     Renovate creates PR immediately once it comes across a new version and then this PR doesn't merge until next 3 days to 
     check the stability.
     It is a noise in the Github repository with a lot of PR's.

**Note**: We still have a drawback using Batching, Because if any single update fails, The PR will not merge and we have to manually check the PR. - Still I prefer this way, instead of having a lot of PR's.
May be check this issue for more details:  https://github.com/renovatebot/renovate/issues/1342

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,11 +5,5 @@
   "automerge": true,
   "schedule": ["every weekend"],
   "ignoreDeps": ["nock"],
-  "prConcurrentLimit": 10,
-  "packageRules": [
-    {
-      "matchDatasources": ["npm"],
-      "stabilityDays": 3
-    }
-  ]
+  "prConcurrentLimit": 10
 }
--- a/renovate.json
+++ b/renovate.json
@@ -4,7 +4,6 @@
   ],
   "automerge": true,
   "schedule": ["every weekend"],
-  "ignoreDeps": ["nock"],
   "groupName": "all",
   "docker": {
     "enabled":  false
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -5,11 +5,9 @@
   "automerge": true,
   "schedule": ["every weekend"],
   "ignoreDeps": ["nock"],
```
lojzatran, line 7: Why nock is ignored?

lojzatran (commented):
I had a look at the issue you posted in the description. Is it done like the guy suggested in 3rd option in the issue?

renovate.json
praveenkumarct: We discussed this already. nock was already ignored in dependabot. So I kept it here.

lojzatran: Yes, I remember you wrote sth like that, but what is the reason it was ignored? I cannot find a reason.

praveenkumarct: Yeah, Even I am not sure exactly on the reason. @ahmetoz do you have any idea on this?

ahmetoz: Don't know the commit is from dependabot and we can not access the old config anymore. the deps related to test not the source code so I would suggest removing that ignore from the config, so we could see the real reason when it occurs again 😉

lojzatran (approved):

PR #806, opened 2021-08-23
Fix for validator-builder.js

peniakoff:
`amountInMakePaymentRequest` should be convert to Number before checking, because it comes from parsed string.
It is a fix for error:
```
{
        "code": "InvalidField",
        "message": "amountPlanned field must be the same as the amount in makePaymentRequest in the interface interactions or makePaymentRequest  in the custom field"
}
```
this PR is related to my opened issue: https://github.com/commercetools/commercetools-adyen-integration/issues/807

PR #805, opened 2021-08-23
Update HowToRun.md

jhilden:
We ran into the issue where `legacyApiBaseUrl` was not set and most API calls were working but some (e.g. manualCapture) were not. It's a bit confusing that this field is not required.

PR #812, opened 2021-08-30
Fix for validator-builder.js

lojzatran:
Original PR: https://github.com/commercetools/commercetools-adyen-integration/pull/806


extension/docs/HowToRun.md
```diff
@@ -73,8 +73,8 @@ Multiple child attributes can be provided in the `adyen` attribute. Each direct
...
 | `apiKey`           | You'll be making API requests that are authenticated with an [API key](https://docs.adyen.com/user-management/how-to-get-the-api-key#page-introduction). | YES      |                                                      |
-| `apiBaseUrl`       | [Checkout endpoint](https://docs.adyen.com/development-resources/live-endpoints#checkout-endpoints) of Adyen.                                            | NO       | `https://checkout-test.adyen.com/v52`                |
-| `legacyApiBaseUrl` | [Standart payment endpoint](https://docs.adyen.com/development-resources/live-endpoints#standard-payments-endpoints) of Adyen.                           | NO       | `https://pal-test.adyen.com/pal/servlet/Payment/v52` |
+| `apiBaseUrl`       | [Checkout endpoint](https://docs.adyen.com/development-resources/live-endpoints#checkout-endpoints) of Adyen.                                            | NO       | `https://checkout-test.adyen.com/v52` (even though it is not required, you **need** to spcefiy a URL for live environment)             |
```
lojzatran: @ahmetoz @leungkinghin @butenkor WDYT about this change? We had an issue that it is not clear enough that for production it needs another value. See: https://github.com/commercetools/commercetools-adyen-integration/pull/805.
Right now we have only in the column name written `Default value (only for test environment)`. Do you think this change is good enough or sth more needs to be done here?

ahmetoz: The PR is a bit confusing, I guess you are mixing different fork PRs. 
I think it's enough with the new documentation update.

lojzatran: Yes, just check this part only. The other changes are not related.

Issue #813, opened 2021-09-02
Notification error logging


wiktor-obrebski:
When my notification service got unexpected error as result of handling Adyen request I got following error:
```json
{
  "name": "ctp-adyen-integration-notifications",
  "hostname": "...",
  "pid": ...,
  "level": ...,
  "notification": [
    {
      "eventCode": "AUTHORISATION",
      "eventDate": "2021-09-02T14:36:37+02:00",
      "pspReference": "...",
      "success": "true"
    }
  ],
  "cause": {},
  "msg": "Unexpected exception occurred.",
  "time": "2021-09-02T12:36:37.592Z",
  "v": 0
}
```

What is easy to spot here is that I have almost no info what goes wrong.
The `cause` is `{}`.

https://github.com/commercetools/commercetools-adyen-integration/blob/fdb6c149b8007ef138cff7089e492a822e7fa64a/notification/src/utils/error-utils.js#L23-L26

I believe it's because the `getErrorCause` return entire error object instead of `error.message` or at least `String(error)`.
When parsing the `Error` object is just parsed to `{}`.

Here the error is throwed:
https://github.com/commercetools/commercetools-adyen-integration/blob/master/notification/src/api/notification/notification.controller.js#L34

_The issue here is to improve error handling, not to solve my specific case._

ahmetoz:
Hey @wiktor-obrebski  this is already in our backlog and we are working on it, thanks for reporting the issue to us.

wiktor-obrebski:
Thank you.

ahmetoz:
Thanks again for reporting the issue, it is fixed on the latest release: https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v9.0.3

PR #816, opened 2021-09-06
Handling unexpected errors with VError.

praveenkumarct:
Handle unexpected errors by returning VError.

Related Issue: https://github.com/commercetools/commercetools-adyen-integration/issues/813

### DIFF
--- a/notification/src/utils/error-utils.js
+++ b/notification/src/utils/error-utils.js
@@ -21,11 +21,15 @@ function isRecoverableError(err) {
 }
 
 function getErrorCause(err) {
-
   if (err instanceof VError) {
     const cause = err.cause()
-    return {name: cause.name, message: cause.message, stack: cause.stack,
-      body: cause.body, statusCode: cause.statusCode}
+    return {
+      name: cause.name,
+      message: cause.message,
+      stack: cause.stack,
+      body: cause.body,
+      statusCode: cause.statusCode,
+    }
   }
 
   return err
--- a/notification/src/utils/error-utils.js
+++ b/notification/src/utils/error-utils.js
@@ -1,5 +1,4 @@
 const VError = require('verror')
-const _ = require('lodash')
 
 /*
  * recoverable: notification delivery can be retried by Adyen (return 500)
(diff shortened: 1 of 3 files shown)


notification/src/utils/error-utils.js
```diff
@@ -21,7 +21,17 @@ function isRecoverableError(err) {
...
+    return {
+      name: cause.name,
+      message: cause.message,
+      stack: cause.stack,
+      body: cause.body,
+      statusCode: cause.statusCode,
+    }
+  }
+
   return err
```
ahmetoz, line 35: I'm sure that this will return an empty log on cause field (try with throwing a simple Error object and check out it's logs), with that I could see that Verror cause is returning an instance of the js Error object and bunyan  (logging module) can not log it properly, maybe this could narrow the investigation ?

ahmetoz (changes requested):

notification/src/utils/error-utils.js
ahmetoz: I found the logging issue here 😉 

In default the bunyan can serialize an `Error object` on err field of the log context, here the Error object is passed in a field called `cause` so bunyan is not serializing it to stdout 

see details: https://github.com/trentm/node-bunyan#requirements-for-serializers-functions

So overcome the issue a custom serializer for `cause` field could be added to the logger: 

https://github.com/commercetools/commercetools-adyen-integration/blob/0602d6529b579cf3f634feb403f7f3d8d2ddea0c/notification/src/utils/logger.js#L12-L14

to:
````javascript
serializers: { err: bunyan.stdSerializers.err, cause: bunyan.stdSerializers.err },
`````

Another approach might be renaming cause to err.

NOTE: I have not test all type of errors (verror, normal error and network error) so please test it.

praveenkumarct: Thanks for your time. https://github.com/commercetools/commercetools-adyen-integration/pull/816/commits/c64df4f331040a2782d4246630d76a3c141e2104

ahmetoz (approved):
🎉

PR #817, opened 2021-09-07
Fix boolean check for removeSensitiveData attribute in Config JSON

leungkinghin:


notification/test/unit/config.spec.js
```diff
@@ -33,4 +33,56 @@ describe('::config::', () => {
...
+        logLevel: 'DEBUG',
+        removeSensitiveData: false,
+      })
+      const config = requireUncached('../../src/config/config')
+      expect(config.getModuleConfig()?.removeSensitiveData).to.eql(false)
+    }
+  )
+
+  it(
+    'when removeSensitiveData is set as boolean true in config.js, ' +
```
lojzatran, line 64: What if it is set as String? You also going to test it?

lojzatran (changes requested):

notification/test/unit/config.spec.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/817/commits/b3ee93bd5b59eb314637b2e6655e1d1cce908da3

extension/test/unit/config/config.spec.js
```diff
@@ -314,4 +314,124 @@ describe('::config::', () => {
...
+            apiKey: 'apiKey',
+            clientKey: 'clientKey',
+            legacyApiBaseUrl: 'legacyApiBaseUrl',
+          },
+        },
+        logLevel: 'DEBUG',
+        removeSensitiveData: false,
+      })
+      const config = requireUncached('../../../src/config/config')
+      expect(config.getModuleConfig()?.removeSensitiveData).to.eql(false)
```
lojzatran, line 348:
```suggestion
      expect(config.getModuleConfig().removeSensitiveData).to.eql(false)
```

Remove it everywhere.

lojzatran (approved):

PR #826, opened 2021-09-23
Add timestamp into transaction

leungkinghin:
TBC

### DIFF
--- a/extension/src/paymentHandler/cancel-payment.handler.js
+++ b/extension/src/paymentHandler/cancel-payment.handler.js
@@ -44,6 +44,7 @@ function _createTransactionActions(paymentObject, pspReference) {
   return [
     pU.createChangeTransactionStateAction(transactionId, 'Pending'),
     pU.createChangeTransactionInteractionId(transactionId, pspReference),
+    pU.createChangeTransactionTimestampAction(transactionId),
   ]
 }
 
--- a/extension/src/paymentHandler/manual-capture.handler.js
+++ b/extension/src/paymentHandler/manual-capture.handler.js
@@ -43,6 +43,9 @@ async function execute(paymentObject) {
         response.pspReference
       )
     )
+    actions.push(
+      pU.createChangeTransactionTimestampAction(chargeInitialTransaction.id)
+    )
   }
 
   return {
--- a/extension/src/paymentHandler/payment-utils.js
+++ b/extension/src/paymentHandler/payment-utils.js
@@ -64,7 +64,7 @@ function createChangeTransactionTimestampAction(transactionId) {
   return {
     action: 'changeTransactionTimestamp',
     transactionId,
-    changeTransactionTimestamp: currentTimestamp.toISOString(),
+    timestamp: currentTimestamp.toISOString(),
   }
 }
 
--- a/extension/src/paymentHandler/payment-utils.js
+++ b/extension/src/paymentHandler/payment-utils.js
@@ -59,6 +59,15 @@ function createChangeTransactionStateAction(transactionId, transactionState) {
   }
 }
 
+function createChangeTransactionTimestampAction(transactionId) {
+  const currentTimestamp = new Date()
+  return {
+    action: 'changeTransactionTimestamp',
+    transactionId,
+    changeTransactionTimestamp: currentTimestamp.toISOString(),
+  }
+}
+
 function createSetCustomFieldAction(name, response) {
   return {
     action: 'setCustomField',
@@ -197,6 +206,7 @@ module.exports = {
   listRefundTransactionsInit,
   createAddInterfaceInteractionAction,
   createChangeTransactionStateAction,
+  createChangeTransactionTimestampAction,
   createSetCustomFieldAction,
   createChangeTransactionInteractionId,
   createAddTransactionAction,
--- a/extension/src/paymentHandler/payment-utils.js
+++ b/extension/src/paymentHandler/payment-utils.js
@@ -101,9 +101,11 @@ function createAddTransactionAction({
   currency,
   interactionId,
 }) {
+  const currentTimestamp = new Date().toISOString()
   return {
     action: 'addTransaction',
     transaction: {
+      timestamp: currentTimestamp,
       type,
       amount: {
         currencyCode: currency,
--- a/extension/src/paymentHandler/refund-payment.handler.js
+++ b/extension/src/paymentHandler/refund-payment.handler.js
@@ -47,6 +47,9 @@ async function execute(paymentObject) {
             response.pspReference
           )
         )
+        actions.push(
+          pU.createChangeTransactionTimestampAction(refundTransaction.id)
+        )
       }
     })
   )
--- a/extension/test/e2e/klarna.spec.js
+++ b/extension/test/e2e/klarna.spec.js
@@ -20,7 +20,7 @@ const {
 } = require('../../src/config/constants')
 
 // Flow description: https://docs.adyen.com/payment-methods/klarna/web-component#page-introduction
-describe('::klarnaPayment::', () => {
+describe.skip('::klarnaPayment::', () => {
   let browser
   let ctpClient
   const adyenMerchantAccount = config.getAllAdyenMerchantAccounts()[0]
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -181,13 +181,17 @@ function calculateUpdateActionsForPayment(payment, notification) {
       )
     else if (
       compareTransactionStates(oldTransaction.state, transactionState) > 0
-    )
+    ) {
       updateActions.push(
         getChangeTransactionStateUpdateAction(
           oldTransaction.id,
           transactionState
         )
       )
+      updateActions.push(
+        getChangeTransactionTimestampUpdateAction(oldTransaction.id)
+      )
+    }
   }
   const paymentMethodFromPayment = payment.paymentMethodInfo.method
   const paymentMethodFromNotification = notificationRequestItem.paymentMethod
@@ -267,6 +271,15 @@ function getChangeTransactionStateUpdateAction(
     action: 'changeTransactionState',
     transactionId,
     state: newTransactionState,
+  }
+}
+
+function getChangeTransactionTimestampUpdateAction(transactionId) {
+  const currentTimestamp = new Date()
+  return {
+    action: 'changeTransactionTimestamp',
+    transactionId,
+    timestamp: currentTimestamp.toISOString(),
   }
 }
 
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -312,9 +312,11 @@ function getAddTransactionUpdateAction({
   currency,
   interactionId,
 }) {
+  const currentTimestamp = new Date().toISOString()
   return {
     action: 'addTransaction',
     transaction: {
+      timestamp: currentTimestamp,
       type,
       amount: {
         currencyCode: currency,
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -312,11 +312,9 @@ function getAddTransactionUpdateAction({
   currency,
   interactionId,
 }) {
-  const currentTimestamp = new Date().toISOString()
   return {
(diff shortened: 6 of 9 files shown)


notification/src/handler/notification/notification.handler.js
```diff
@@ -270,6 +274,15 @@ function getChangeTransactionStateUpdateAction(
   }
 }
 
+function getChangeTransactionTimestampUpdateAction(transactionId) {
+  const currentTimestamp = new Date()
+  return {
+    action: 'changeTransactionTimestamp',
+    transactionId,
+    timestamp: currentTimestamp.toISOString(),
```
lojzatran, line 282: In the notification we have a field `eventDate` with timestamp. Why did you choose to create the timestamp in code and not taking it from the notification? https://docs.adyen.com/api-explorer/#/Webhooks/v1/post/REFUND

extension/src/paymentHandler/payment-utils.js
```diff
@@ -59,6 +59,15 @@ function createChangeTransactionStateAction(transactionId, transactionState) {
   }
 }
 
+function createChangeTransactionTimestampAction(transactionId) {
+  const currentTimestamp = new Date()
```
lojzatran, line 63: When we make a request to Adyen, we have in response header the field `date`. Currently we're not using it anyhow. Why did you decide to use timestamp created in code instead?

lojzatran (changes requested):

notification/src/handler/notification/notification.handler.js
leungkinghin: I thought we take actual transaction create/update moment as timestamp
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/63c5e10a57b035d00ab5596aaf966e45d53a3130

lojzatran: I'm just asking why, maybe you're right and we should take the actual timestamp. Did you discuss with other devs about it?

extension/src/paymentHandler/payment-utils.js
leungkinghin: Not sure if it is also applicable for serverless function. Need to see any other `date` attributes provided in `paymentObject`

extension/src/paymentHandler/cancel-payment.handler.js
```diff
@@ -44,6 +44,10 @@ function _createTransactionActions(paymentObject, pspReference) {
   return [
     pU.createChangeTransactionStateAction(transactionId, 'Pending'),
     pU.createChangeTransactionInteractionId(transactionId, pspReference),
+    pU.createChangeTransactionTimestampAction(
+      transactionId,
+      paymentObject.createdAt
```
butenkor, line 49: Since at this point we had already response from Adyen why we take timestamp from the payment object and not from the response or Adyen notification? Also, payment createdAt might be very different from the actual processing of the transaction, right?

extension/src/paymentHandler/manual-capture.handler.js
```diff
@@ -43,6 +43,12 @@ async function execute(paymentObject) {
         response.pspReference
       )
     )
+    actions.push(
+      pU.createChangeTransactionTimestampAction(
+        chargeInitialTransaction.id,
+        paymentObject.createdAt
```
butenkor, line 49: same as above

extension/src/paymentHandler/payment-utils.js
```diff
@@ -59,6 +59,19 @@ function createChangeTransactionStateAction(transactionId, transactionState) {
...
+  transactionId,
+  paymentObjTimestamp
+) {
+  let currentTimestamp = new Date().toISOString()
+  if (paymentObjTimestamp !== undefined) currentTimestamp = paymentObjTimestamp
+  return {
+    action: 'changeTransactionTimestamp',
+    transactionId,
+    timestamp: currentTimestamp,
+  }
```
butenkor, line 72: A bit shorter and maybe a bit easier to read:

```
  return {
    action: 'changeTransactionTimestamp',
    transactionId,
    timestamp: (paymentObjTimestamp == undefined) ? new Date().toISOString() : paymentObjTimestamp,
  }
```

extension/src/paymentHandler/payment-utils.js
```diff
@@ -59,6 +59,19 @@ function createChangeTransactionStateAction(transactionId, transactionState) {
   }
 }
 
+function createChangeTransactionTimestampAction(
+  transactionId,
+  paymentObjTimestamp
```
butenkor, line 64: why do we always assume that it is the "payment object time stamp" and not just "timestamp" since it could be also a timestamp from the Adyen response, right?

extension/src/paymentHandler/refund-payment.handler.js
```diff
@@ -47,6 +47,12 @@ async function execute(paymentObject) {
             response.pspReference
           )
         )
+        actions.push(
+          pU.createChangeTransactionTimestampAction(
+            refundTransaction.id,
+            paymentObject.createdAt
```
butenkor, line 53: same as above

extension/test/e2e/klarna.spec.js
```diff
@@ -20,7 +20,7 @@ const {
 } = require('../../src/config/constants')
 
 // Flow description: https://docs.adyen.com/payment-methods/klarna/web-component#page-introduction
-describe('::klarnaPayment::', () => {
+describe.skip('::klarnaPayment::', () => {
```
butenkor, line 23: why?

notification/src/handler/notification/notification.handler.js
```diff
@@ -181,13 +181,20 @@ function calculateUpdateActionsForPayment(payment, notification) {
...
       updateActions.push(
         getChangeTransactionStateUpdateAction(
           oldTransaction.id,
           transactionState
         )
       )
+      updateActions.push(
+        getChangeTransactionTimestampUpdateAction(
+          oldTransaction.id,
+          notificationRequestItem.eventDate
```
butenkor, line 194: Is the timezone of the notification always the same or depends on Adyen account? I wonder if we should use the same timezone or always convert to UTC as all data in commercetools?

notification/src/handler/notification/notification.handler.js
```diff
@@ -270,6 +277,29 @@ function getChangeTransactionStateUpdateAction(
...
+  transactionEventDate
+) {
+  let transactionTimestamp = ''
+  try {
+    // Assume transactionEventDate should be in correct format (e.g. '2019-01-30T18:16:22+01:00')
+    const eventDateMilliSecondsStr = Date.parse(transactionEventDate)
+    const transactionDate = new Date()
+    transactionDate.setTime(eventDateMilliSecondsStr)
+    transactionTimestamp = transactionDate.toISOString()
+  } catch (err) {
```
butenkor, line 291: We just cache and log nothing? In case the Adyen format is always wrong we would never be able to use Adyen date and would not even notice it?

notification/src/handler/notification/notification.handler.js
```diff
@@ -312,9 +342,11 @@ function getAddTransactionUpdateAction({
   currency,
   interactionId,
 }) {
+  const currentTimestamp = new Date().toISOString()
```
butenkor, line 345: Why we generate instead of taking it from Adyen notification?

extension/src/paymentHandler/cancel-payment.handler.js
lojzatran: Actually at this point we don't have the timestamp from notification. We will have a transaction with Pending state. I would put the timestamp there like Hin did here so that we know when the transaction was added. It could be that the transaction will stay in Pending for some time.

extension/src/paymentHandler/payment-utils.js
leungkinghin: As discuss, we don't need to set timestamp in extension module explicitly. We just need the event date returned from notification module.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/9e35ab7f3f1285abb7119ef3a73d820293b3fba2

extension/src/paymentHandler/cancel-payment.handler.js
leungkinghin: As discuss, we don't need to set timestamp in extension module explicitly. We just need the event date returned from notification module.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/9e35ab7f3f1285abb7119ef3a73d820293b3fba2

extension/src/paymentHandler/manual-capture.handler.js
leungkinghin: Undo the timestamp assignment in extension module.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/9e35ab7f3f1285abb7119ef3a73d820293b3fba2

extension/src/paymentHandler/payment-utils.js
leungkinghin: Undo the timestamp assignment in extension module.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/9e35ab7f3f1285abb7119ef3a73d820293b3fba2

extension/src/paymentHandler/payment-utils.js
leungkinghin: Undo the timestamp assignment in extension module.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/9e35ab7f3f1285abb7119ef3a73d820293b3fba2

extension/src/paymentHandler/refund-payment.handler.js
leungkinghin: Undo the timestamp assignment in extension module.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/9e35ab7f3f1285abb7119ef3a73d820293b3fba2

notification/src/handler/notification/notification.handler.js
leungkinghin: Get `eventDate` from notification and apply it in `addTransaction` update action
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/8dc3daa1e3fd8001d933c30e260cdd3794e675be

notification/src/handler/notification/notification.handler.js
leungkinghin: Add error logging when conversion failed
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/a0187be017ff3979aba687198953b0738085aea0

notification/src/handler/notification/notification.handler.js
leungkinghin: Commercetools only accepts UTC representation for transaction timestamp.
https://docs.commercetools.com/api/projects/payments#transaction
https://docs.commercetools.com/api/types#datetime

I don't think we can use same timezone as notification response.

extension/test/e2e/klarna.spec.js
leungkinghin: It failed always before, but now it has been fixed in master branch.
https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/fe6356403e1c4e8ec1320ad845ea68d6dcc06bed

notification/src/handler/notification/notification.handler.js
```diff
@@ -270,6 +281,35 @@ function getChangeTransactionStateUpdateAction(
   }
 }
 
+function convertDateToUTCFormat(transactionEventDate, logger) {
+  try {
+    // Assume transactionEventDate should be in correct format (e.g. '2019-01-30T18:16:22+01:00')
+    const eventDateMilliSecondsStr = Date.parse(transactionEventDate)
+    const transactionDate = new Date()
```
lojzatran, line 288:
```suggestion
    const transactionDate = new Date(eventDateMilliSecondsStr)
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters

lojzatran (approved):

butenkor (approved):

notification/src/handler/notification/notification.handler.js
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/826/commits/da4e9bc8a0ac09719dbc657d3d9fee96b4d09a38

PR #837, opened 2021-10-05
Avoid cleanup payment after test

leungkinghin:
**Situation**
Avoid cleaning-up payment after test so that it is easier for developer to examine the result.

**Hints for developer**
- No need to clean up resource before and after each integration test case.
- Modified `package.json`. It aims at avoiding to restart local server between integration test and e2e test, which removes sources (incl. payment) before e2e test cases execution.


extension/package.json
```diff
@@ -5,11 +5,13 @@
   "license": "MIT",
   "scripts": {
     "check-coverage": "nyc check-coverage --statements 97",
-    "test": "npm run check && npm run cover:unit && npm run cover:integration && npm run cover:e2e && npm run check-coverage",
+    "test": "npm run check && npm run cover:unit && npm run cover:integration-plus-e2e && npm run check-coverage",
```
lojzatran, line 8: I don't understand this change. Why you need to make it like this? I thought we don't clean up any resources anymore?

lojzatran (changes requested):
What about the notification module?

extension/package.json
leungkinghin: Keep payments and types all the time when test server is started.
https://github.com/commercetools/commercetools-adyen-integration/pull/837/commits/1c9c2e4b6290b8ea7b3c49ce0d0fdb9f9cf5ce81

ahmetoz:
Redundant with #890 #891

PR #844, opened 2021-10-15
Upgrade adyen api version

lojzatran:
TODO: fix failing tests


lojzatran:
This PR won't be done as the newest version of Adyen frontend JS there is a different flow. We updated the backend API to the newest version already: https://github.com/commercetools/commercetools-adyen-integration/pull/852

PR #854, opened 2021-11-05
add get-carbon-offset-costs.handler and cover it with unit tests

ahmetoz:
Overview: https://www.adyen.com/social-responsibility/impact#restore

Concept: Offset information will be requested through our adyen integration, integration user will set a custom field and we will call Adyen API and set it back to response custom field. This needs to be before the makePaymentRequest. 

PR addresses the offset calculation only. Integration tests and documentation will be added part of another task.

### DIFF
--- a/extension/src/config/constants.js
+++ b/extension/src/config/constants.js
@@ -23,4 +23,6 @@ module.exports = {
   ],
 
   PAYMENT_METHOD_TYPE_AFFIRM_METHODS: ['affirm'],
+  CTP_CARBON_OFFSET_COSTS_RESPONSE: 'getCarbonOffsetCostsResponse',
+  CTP_INTERACTION_TYPE_GET_CARBON_OFFSET_COSTS: 'getCarbonOffsetCosts',
 }
--- a/extension/src/paymentHandler/payment-handler.js
+++ b/extension/src/paymentHandler/payment-handler.js
@@ -6,6 +6,7 @@ const submitPaymentDetailsHandler = require('./submit-payment-details.handler')
 const manualCaptureHandler = require('./manual-capture.handler')
 const cancelHandler = require('./cancel-payment.handler')
 const refundHandler = require('./refund-payment.handler')
+const getCarbonOffsetCostsHandler = require('./get-carbon-offset-costs.handler')
 const pU = require('./payment-utils')
 const auth = require('../validator/authentication')
 const errorMessages = require('../validator/error-messages')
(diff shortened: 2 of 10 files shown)


extension/src/paymentHandler/payment-handler.js
```diff
@@ -77,6 +78,11 @@ function _getPaymentHandlers(paymentObject) {
     !paymentObject.custom.fields.getPaymentMethodsResponse
   )
     handlers.push(getPaymentMethodsHandler)
+  if (
+    paymentObject.custom.fields.getCarbonOffsetCostsRequest &&
+    !paymentObject.custom.fields.getCarbonOffsetCostsResponse
```
butenkor, line 83: I wonder if we should validate if request is done after makePaymentResponse - which does not make much sense anymore?

lojzatran (approved):

extension/src/config/constants.js
```diff
@@ -23,4 +23,6 @@ module.exports = {
   ],
 
   PAYMENT_METHOD_TYPE_AFFIRM_METHODS: ['affirm'],
+  CTP_CARBON_OFFSET_COSTS_RESPONSE: 'getCarbonOffsetCostsResponse',
+  CTP_INTERACTION_TYPE_GET_CARBON_OFFSET_COSTS: 'getCarbonOffsetCosts',
```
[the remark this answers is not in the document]
ahmetoz: this is pattern for interface interaction custom type same for other requests (e.g: makePayment) so I don't want to change it. it includes request and response fields together and this is the name of type. For example: 

````json
"interfaceInteractions": [
        {
          "type": {
            "typeId": "type",
            "id": "812889c7-e212-4d30-908f-90d0d783f0ed"
          },
          "fields": {
            "createdAt": "2021-11-11T20:20:36.114Z",
            "response": "{\"deliveryOffset\":{\"currency\":\"EUR\",\"value\":0},\"productOffset\":{\"currency\":\"EUR\",\"value\":138},\"totalOffset\":{\"currency\":\"EUR\",\"value\":138},\"unavailableProductCodes\":[\"123\"]}",
            "request": "{\"originCountry\":\"BE\",\"deliveryCountry\":\"FR\",\"packageWeight\":{\"value\":2.2,\"unit\":\"kg\"},\"products\":[{\"code\":\"123\",\"weight\":{\"value\":0.2,\"unit\":\"kg\"}},{\"code\":\"10001335\",\"weight\":{\"value\":2,\"unit\":\"kg\"}}],\"merchantAccount\":\"CommercetoolsGmbHDE775\"}",
            "type": "getCarbonOffsetCosts"
          }
        }
      ]
````

extension/src/paymentHandler/payment-handler.js
ahmetoz: It's same for getPaymentMethodsRequest. We could put additional restriction here but I don't know if the integration users writes implementations with undocumented patterns 🤔 

@butenkor @lojzatran what do you think should I add a restriction for those that those could be used only before makePayment.

PR #868, opened 2021-11-20
Disable test that has changed the flow

lojzatran:
I had a quick look and the only test that is constantly failing is extension/test/e2e/credit-card-3ds-native.spec.js with MasterCard.

The issue is the payment verification page for MasterCard 3D has changed and is now different from before. Therefore our selectors do not work anymore.

After disabling MasterCard tests are working again.

Suggestion: I would remove completely MasterCard from the 3D verification test and leave Visa there. For testing the flow it's enough. Having another credit card do not add much value comparing to the maintenance we have to do. WDYT?

### DIFF
--- a/extension/test/e2e/credit-card-3ds-native.spec.js
+++ b/extension/test/e2e/credit-card-3ds-native.spec.js
@@ -21,14 +21,6 @@ describe('::creditCardPayment3dsNative::', () => {
   const adyenMerchantAccount = config.getAllAdyenMerchantAccounts()[0]
   const ctpProjectKey = config.getAllCtpProjectKeys()[0]
 
-  // See more: https://docs.adyen.com/development-resources/test-cards/test-card-numbers
-  const creditCards = [
-    {
-      name: 'Visa',
-      creditCardNumber: '4917 6100 0000 0000',
-    },
-  ]
-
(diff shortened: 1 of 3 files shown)


extension/test/e2e/credit-card-3ds-native.spec.js
```diff
@@ -182,8 +182,6 @@ describe('::creditCardPayment3dsNative::', () => {
       makePaymentResponse
     )
 
-    await browserTab.waitForTimeout(15_000)
```
lojzatran, line 185: I removed this as in the next line in the method `await creditCardNativePage.finish3dsNativePayment()` we wait again. So we had 2x waitForTimeout, which is redundant.

ahmetoz (approved):
Kudos for finding the issue 👍🏽

lojzatran:
> Kudos for finding the issue 👍🏽

Thanks :) so should I remove the MasterCard test? WDYT?

ahmetoz:
> > Kudos for finding the issue 👍🏽
> 
> Thanks :) so should I remove the MasterCard test? WDYT?

Yes, I think it's flaky and we are covering the main functionality already 👍🏽

Issue #881, opened 2021-12-13
Question: Adding metadata


BielsasBucket:
Hi, not sure this is an issue. More a question.

Is there a way we can attach more metadata to the adyen request using the extension?

Tried looking in documentation but couldn't see anything. Will check again tho.

SDSellers:
We would like the ability to attach additional metadata to help with internal reporting and analytics. 

For Web Payments, we are still defining what metadata we want but we need the ability to add some.

Our project which was for Store Payments (i.e. Adyen's Terminal API) we attached the ctPaymentId, Shop Assistant Name, Order Number, Receipt Number, StoreId. etc - this helps our data and finance team with reporting.

I hope this answers your query and it would be great if this could be done soon.

Many Thanks.

ahmetoz:
Hey @SDSellers thanks for describing the use case, currently we are evaluating to extend metadata object passed through with custom fields, so yes we would add it as part of the integration so you could pass your metadata with extension module/service. 
Does this help ?

SDSellers:
Yes, it does. Thank you.

One more question, roughly when would this new feature for passing metadata be released. Its due to the nature of my current project so we can plan.

ahmetoz:
@BielsasBucket it will be released soon, I'm working on it.

BielsasBucket:
Thanks, I'll close this for now then.

PR #883, opened 2021-12-15
Make metadata extendible

ahmetoz:
Addresses: #881 

Example notification payload, (see metadata fields on additionalData):

````json
{
  "NotificationRequestItem": {
    "additionalData": {
      "cvcResult": "1 Matches",
      "expiryDate": "03/2030",
      "metadata.orderNumber": "externalOrderSystem-12345",
      "authCode": "094419",
      "avsResult": "5 No AVS data provided",
      "cardSummary": "1111",
      "metadata.ctProjectKey": "test",
      "authorisationMid": "1000",
      "hmacSignature": "Z/v",
      "acquirerAccountCode": "TestPmmAcquirerAccount",
      "metadata.receiptNumber": "externalOrderSystem-receipt123"
    },
    "amount": {
      "currency": "EUR",
      "value": 1000
    },
    "eventCode": "AUTHORISATION",
    "eventDate": "2021-12-15T09:58:29+01:00",
    "merchantAccountCode": "Adyen-ctp-qa-01",
    "merchantReference": "dfdsfsdfsdc",
    "operations": [
      "CANCEL",
      "CAPTURE",
      "REFUND"
    ],
    "paymentMethod": "visa",
    "pspReference": "852639558709915E",
    "reason": "0000:1111:03/2030",
    "success": "true"
  }
}
````

### DIFF
--- a/extension/src/service/web-component-service.js
+++ b/extension/src/service/web-component-service.js
@@ -116,7 +116,7 @@ function extendRequestObjWithMetadata(requestObj, commercetoolsProjectKey) {
   if (requestObj.metadata) {
     requestObj.metadata = {
       ...requestObj.metadata,
-      ctProjectKey: commercetoolsProjectKey
+      ctProjectKey: commercetoolsProjectKey,
     }
   } else {
     requestObj.metadata = {
--- a/extension/src/service/web-component-service.js
+++ b/extension/src/service/web-component-service.js
@@ -113,10 +113,17 @@ function extendRequestObjWithApplicationInfo(requestObj) {
 }
 
 function extendRequestObjWithMetadata(requestObj, commercetoolsProjectKey) {
-  requestObj.metadata = {
-    // metadata key must have length of max. 20 chars
-    // metadata value must have length of max. 80 chars
-    ctProjectKey: commercetoolsProjectKey,
+  if (requestObj.metadata) {
+    requestObj.metadata = {
+      ...requestObj.metadata,
+      ctProjectKey: commercetoolsProjectKey
+    }
+  } else {
+    requestObj.metadata = {
+      // metadata key must have length of max. 20 chars
+      // metadata value must have length of max. 80 chars
+      ctProjectKey: commercetoolsProjectKey,
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ahmetoz** [comment]
Note for CI: creditCardPayment3dsNative e2e test seems flaky, will check soon.

File: extension/test/integration/make-payment.handler.spec.js [hunk 1]
```diff
@@ -95,7 +100,7 @@ describe('::make-payment with multiple adyen accounts use case::', () => {
         interaction.fields.type === constants.CTP_INTERACTION_TYPE_MAKE_PAYMENT
     )
     const makePaymentRequest = JSON.parse(interfaceInteraction.fields.request)
-    expect(makePaymentRequest.metadata).to.deep.equal({
+    expect(makePaymentRequest.metadata).to.deep.include({
```
lojzatran, line 103: Can you check that it contains also other metadata?

extension/src/service/web-component-service.js
```diff
@@ -113,10 +113,17 @@ function extendRequestObjWithApplicationInfo(requestObj) {
 }
 
 function extendRequestObjWithMetadata(requestObj, commercetoolsProjectKey) {
-  requestObj.metadata = {
-    // metadata key must have length of max. 20 chars
-    // metadata value must have length of max. 80 chars
-    ctProjectKey: commercetoolsProjectKey,
+  if (requestObj.metadata) {
+    requestObj.metadata = {
+      ...requestObj.metadata,
```
lojzatran, line 118: It is allowed to pass only 20 metadata: https://docs.adyen.com/api-explorer/#/CheckoutService/v68/post/payments__reqParam_metadata. Should we already make a check here and reject the request immediately?

lojzatran (changes requested):

extension/src/service/web-component-service.js
butenkor: I would not validate - we can not handle all the possible Adyen errors in the integration. We just need to make sure that our generated payload is valid but user specific data should be validated by Adyen unless it poses an issue for the integration itself, which is not the case here, right?

butenkor (approved):

extension/test/integration/make-payment.handler.spec.js
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/883/commits/001697eb49d7d950cbe254e37b1aab7ff4fc5ec7

lojzatran (approved):

PR #884, opened 2021-12-15
doc: change to notification

danilomsou:
Documentation change, extension was written where notification should be.


ahmetoz:
@danilomsou thanks for the fix !

PR #890, opened 2021-12-21
Replace local tunnel with automated google function deployment (only fo CI)

ahmetoz:
**Situation**
Local tunnel is used to exposes the localhost to internet for testing purposes as API extension trigger url expects a valid URL to trigger.
Currently it's used on the integration and e2e tests, and creating a tunnel before the tests started. See: https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/test/integration/integration-test-set-up.js#L65-L80

**Complication**
Local tunnel is working well for development purposes (e.g debugging) but does not simulate the reality of the integration. As extension is running on a local node server, but it should be running either in a container or in a serverless function such as google functions. Also we faced some issues with localtunnel on tests that is being flaky (resolved but might cause another issue on the future)

**Resolution**
Replace local tunnel such as using google function function URL as an API extension trigger url on test fixtures.

------

# Review notes: 
- First challange was doing automated deployment to GCP, it's resolved with gcloud sdk (see: ci.yaml changes) 
- Second challange was not removing resources before and after the tests to make the test scenarios more realistic.
- Local tunnel is useful for development (debugging purpose) so it's the default hook if the CI env variable is not set ( github actions sets this variable to true so CI flow uses GCP functions on test fixtures)

PR #891, opened 2021-12-22
Avoid cleanup resources on notification test hooks and use GCP function on CI.

ahmetoz:
Related PR with #890 focusing on the notification part.

#### Review notes:
- Overall CI build time is increased around 4 minutes because of GCP function deployment, now we have around 6-7 minutes on build.
- On debugging notification an extension mechanism is needed on updating/creating payments but this is hard on development so it sets up a tunnel which will be triggered by project extension but tunnel will not use real extension server  that tunnelled api extension will just return 200 so the payment is created/updated without an issue. (see: 788d11e)
- The hard coded configurations are changed to dummy values so the tests always react to given configuration dynamically. (see: 9da5323)
- Coverage challange on CI flow: before the tests were passing through the node server instead of real serverless function on CI, but now they are flowing through function, which resulted that the server part of the source code is not covered, we decided to keep it as is instead of writing more unit tests only for the server parts.


notification/package.json
```diff
@@ -3,12 +3,12 @@
   "version": "9.4.0",
   "description": "Part of the integration of Adyen with commercetools responsible to receive and process notifications from Adyen",
   "scripts": {
-    "check-coverage": "nyc check-coverage --statements 93",
+    "check-coverage": "nyc check-coverage --lines 83",
```
lojzatran, line 6: Is it such a big drop of 10% when we go serverless? What's the diff between statements and lines?

lojzatran (approved):
Good job!

notification/package.json
ahmetoz: Previous (master)
File                                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------------------------------|---------|----------|---------|---------|-------------------
All files                              |   93.78 |    81.43 |   98.61 |   94.21 | 

this branch:
File                                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------------------------------|---------|----------|---------|---------|-------------------
All files                              |   83.42 |    71.83 |   77.03 |    84.4 |

PR #892, opened 2021-12-22
do not create unnecessary payment requests with e2e test fixtures

ahmetoz:
**Situation**
In end-to-end test, we execute different process such as create payment, make payment, handle redirect, payment authentication, etc. in order to simulate the entire payment process for testing.

**Complication**
Payment creation without makePaymentRequest triggers extension module unnecessarily and makes the extension to trigger twice: on payment create and payment update with makePaymentRequest.

**Resolution**
Make sure we create payment when makePaymentRequest is available too and thus avoid unnecessary API requests.


lojzatran (commented):
In the integration tests in affirm-make-payment.spec.js and klarna-make-payment.spec.js we are also doing the same ting. If you want, you can also remove the create requests from there and have create and makePayment at the same time.

ahmetoz:
> In the integration tests in affirm-make-payment.spec.js and klarna-make-payment.spec.js we are also doing the same ting. If you want, you can also remove the create requests from there and have create and makePayment at the same time.

Hey @lojzatran the code below check if the cart has paymentId, so in this case the cart needs to have a payment referenced to cart to be able to prepare it's lineItems. Do you see a problem or workaround to be able to that ? I guess in checkout implementation the cart needs to have this payment referenced to it before the makePaymentRequest, right ?

https://github.com/commercetools/commercetools-adyen-integration/blob/02d968c56982e56a952339cdefc82874cf6c528f/extension/src/paymentHandler/make-lineitems-payment.handler.js#L35-L38

lojzatran:
> > In the integration tests in affirm-make-payment.spec.js and klarna-make-payment.spec.js we are also doing the same ting. If you want, you can also remove the create requests from there and have create and makePayment at the same time.
> 
> Hey @lojzatran the code below check if the cart has paymentId, so in this case the cart needs to have a payment referenced to cart to be able to prepare it's lineItems. Do you see a problem or workaround to be able to do that with one request ? I guess in checkout implementation the cart needs to have this payment referenced to it before the makePaymentRequest, right ?
> 
> https://github.com/commercetools/commercetools-adyen-integration/blob/02d968c56982e56a952339cdefc82874cf6c528f/extension/src/paymentHandler/make-lineitems-payment.handler.js#L35-L38

You're right, we can leave it as it is now.

lojzatran (approved):

PR #893, opened 2021-12-24
Support does not need to be contacted anymore, 10 sec is the new limit for payment extensions

cneijenhuis:
See https://docs.commercetools.com/api/releases/2021-12-09-increased-maximum-timeout-for-payment-related-api-extensions


lojzatran (approved):

lojzatran:
Thank you for contribution!

PR #897, opened 2021-12-30
Fix 3DS native test

lojzatran:
The problems:

- 3DS native now requires clientKey, so I added it. 

- Additionally somehow the test could not click on one button in the iframe anymore. I had to update the clicking mechanism to make it work.

- Affirm seems to be working fine, so there is no fix needed.


ahmetoz (approved):
💯

Issue #902, opened 2022-01-10
Problems with refund


wiktor-obrebski:
Hi.
I have a problem with refund feature that makes me doubt how it should really works. 
In [your documentation](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/Refund.md#prerequisites) you clearly state:

> It is required that the payment has one Charge transaction with state Success. If Charge transaction with state Success is NOT present, then it is required that the payment has one Authorized transaction with state Success. From either Charge or Authorized transaction, the interactionId field is being used as originalReference for the Adyen refund request.

And this is how your code works, it's first try to find "Charge" transaction in the payment, and fallback to "Authorization" transaction.
https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/src/paymentHandler/refund-payment.handler.js#L7

But the problem is that [Adyen refund endpoint](https://docs.adyen.com/api-explorer/#/Payment/v67/post/refund__reqParam_originalReference) do not expect "Charge" (in Adyen nomenclature called "Capture")  notification pspRefernces. It's accept only references from:
 - authorisation response
 - authorisation notification

In my tests are it's the case - it's work correctly if I use the Adyen payment `pspReference` (stored in CT `Authorization` transaction), but do not work with the "Capture" `pspReference` stored in "Charge" transaction.

I found and when working with your service, but for faster tests you can forge the requests by yourself (you just need a payment in Adyen with "Capture" notification).
Url: `https://pal-test.adyen.com/pal/servlet/Payment/v67/refund`
POST body:
```json
{
  "originalReference": "...",
  "modificationAmount": {
    "value": 123,
    "currency": "EUR"
  },
  "merchantAccount": "..."
}
```

My tests: it works correctly when I use `pspReference` from CT `Authorisation` transaction, but do not when I use `pspReference` from `Charge` transaction. In second scenario I getting notification with error reason `Transaction not found` after some time.

What I missing? All other features of `commercetools-adyen-integration`  (including capture) work for me with current configuration.

lojzatran:
Hi @wiktor-obrebski,

Thank you for reporting this issue. I had a quick test and I also got an error in the notification like you have. I will test a bit more tomorrow and discuss with the team. I will get back to you with more details.

lojzatran:
Hi @wiktor-obrebski,

I created a PR with the fix. Once it will be approved by another dev, we will release it (most likely within this week).

wiktor-obrebski:
[...]

PR #903, opened 2022-01-11
Remove "charge" transactions from refund process

lojzatran:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/902

### Description
In our code we were taking the pspReference in Charge transactions (mapped as interactionID) when doing Refund (https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/src/paymentHandler/refund-payment.handler.js#L7). However, according to the [docs of Adyen](https://docs.adyen.com/api-explorer/#/Payment/v67/post/refund__reqParam_originalReference), pspReference should be always taken from Authorization transactions only. I tested it and for Charge transactions the returned notifications were not successful. This PR should fix this problem.


extension/docs/Refund.md
```diff
@@ -20,9 +20,8 @@ If you want to return the funds to your shopper, for example if they returned an
 
 #### Prerequisites
 
-It is required that the payment has one `Charge` transaction with state `Success`.
-If `Charge` transaction with state `Success` is NOT present, then it is required that the payment has one `Authorized` transaction with state `Success`.
-From either `Charge` or `Authorized` transaction, the `interactionId` field is being used as `originalReference` for the Adyen refund request.
+It is required that the payment has one `Authorized` transaction with state `Success`.
```
butenkor, line 23:
```suggestion
It is required that the payment has a transaction of type `Authorization` and state `Success`.
```

ahmetoz (approved):
💯

PR #917, opened 2022-02-01
combine faq and best practices readme/markdown docs into one doc as faq doc.

ahmetoz:


docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+Yes, in case you encounter errors in your integration, refer to the [error-handling section](../extension/docs/WebComponentsIntegrationGuide.md#error-handling) on the integration documentation.
+
+### When we should create commercetools order ? 
+
+There are 2 approaches for creating commercetools order, one is creating order after a successful payment, and other is creating before the successful payment.
+Both approaches have their good and bad sides, but we found out that creating an order AFTER a successful payment is less harmful. Please read the details below for both approaches and decide based on your checkout scenarios.
+
+#### What to consider when creating order _AFTER_ a successful commercetools payment ?
+
+- **Shop was not reachable due to network issues**: It is possible that shop was not reachable due to network issues. In that case there will be a successful payment but no order.
```
butenkor, line 98:
```suggestion
- **Shop (success  redirect URL) is not reachable due to the network issues**: Since shop creates an order and  success shop redirect URL can not be reached we might end up with a successful payment but no order.
```

docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+  - **Possible solution**: Create an order based on the payment transaction changes, which delivery is guaranteed due to the asynchronous notifications from Adyen. Depending on your preference you might either query for the latest messages of type [PaymentTransactionAdded](https://docs.commercetools.com/api/message-types#paymenttransactionadded-message), [PaymentTransactionStateChanged](https://docs.commercetools.com/api/message-types#paymenttransactionstatechanged-message) or [subscribe](https://docs.commercetools.com/api/projects/subscriptions#create-a-subscription) to the mentioned message types. Every [message](https://docs.commercetools.com/api/message-types#message) will link its payment through the `resource` field and since every payment is attached to a cart one has all the informations at hand to decide if the cart has to be converted to order or not. Since the job or worker processing the message is not a usual place where the order is created it might be reasonable to pass the cart ID to another service or web shop URL which will verify the cart and create an order out of it.
```
butenkor, line 99:
```suggestion
  - **Possible solution**: Create an order asynchronously based on the payment transaction changes, which delivery is guaranteed due to the asynchronous notifications from Adyen. Depending on your preference you might either query for the latest messages of type [PaymentTransactionAdded](https://docs.commercetools.com/api/message-types#paymenttransactionadded-message), [PaymentTransactionStateChanged](https://docs.commercetools.com/api/message-types#paymenttransactionstatechanged-message) or [subscribe](https://docs.commercetools.com/api/projects/subscriptions#create-a-subscription) to the mentioned message types. Every [message](https://docs.commercetools.com/api/message-types#message) will link its payment through the `resource` field and since every payment is attached to a cart one has all the informations at hand to decide if the cart has to be converted to order or not. Since the job or worker processing the message is not a usual place where the order is created it might be reasonable to pass the cart ID to another service or web shop URL which will verify the cart and create an order out of it.
```

docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+- **More than 1 valid payments on the cart/order**: It is possible to have more than 1 valid payments on the cart/order. During checkout in two tabs, two redirect payments
```
butenkor, line 100:
```suggestion
- **More than 1 successful payments on the cart/order**: It is possible to have more than 1 valid payments on the cart/order. It could happen for example when customer initiates a payment in two different tabs for the same cart and both payments are of type redirect (like credit card and paypal.
```

docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+  (like credit card and paypal) can be created for the same cart. These two payments can be completed independently in both tabs. Since every payment should be always attached to the cart this would make a cart to link two successful payments.
```
butenkor, line 101:
```suggestion
 These two payments can be completed independently in both tabs. Since every payment should be always attached to the cart this would make a cart to link two successful payments.
```

docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+  - **Possible solution**: [Refund](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/docs/Refund.md) one of the successful payments. Similar as in case of `Create an order based on transaction state changes` above one could process the same message types in order to figure out if the cart has to many successful payments and create a refund.  
+- **The amount of the successful payment is lower than the cart amount**: During the checkout the shopper might navigate with two tabs, one with the cart and the other tab might be already a redirected payment with fixed amount like PayPal. Shopper can add more items to the cart so that the cart value is not equal to the  amount that will be paid in the payment provider tab.
```
butenkor, line 103:
```suggestion
- **The amount of the successful payment is lower than the cart amount**: During the checkout the shopper might navigate with two tabs, one with the cart and the other tab might be already a redirected payment with fixed amount like PayPal. Shopper can add more items to the cart so that the cart value is not equal to the amount that will be paid in the payment provider tab.
```

docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+  - **Possible solution**: After a successful payment, validate the cart to check if the cart amount matches payment amount. If not, refund/cancel the payment and ask the shopper to pay again.
```
butenkor, line 104:
```suggestion
  - **Possible solution**: After a successful payment, validate the cart to check if the cart amount matches the paid amount. If not, refund/cancel the payment and ask the shopper to pay again.
```

butenkor (changes requested):

docs/FAQ.md
```diff
@@ -35,3 +60,54 @@ It first find the payment by `key` where `key=${merchantReference}` and then it
...
+- To protect your notification service from unauthorized notifications, we strongly recommend that you activate Hash-based message authentication code [HMAC signatures](../notification/docs/IntegrationGuide.md#step-1-set-up-notification-webhook-and-generate-hmac-signature) during the notification setup.
+
+### What is the best practices for creating/maintaining commercetools payment object ?
+
+- Create your commercetools payment with following the [creating commercetools payment guide](../extension/docs/WebComponentsIntegrationGuide.md#step-2-creating-a-commercetools-payment).
+- After successful payment creation always [add](https://docs.commercetools.com/api/projects/carts#add-payment) it to the appropriate cart.
+- **Never delete or un-assign** created payment objects during the checkout from the cart.
+- Currently [payment status](https://docs.commercetools.com/api/projects/payments#paymentstatus) is not maintained by the integration as status of the payment can be derived from the state of the payment transaction(s).
```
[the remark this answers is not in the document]
ahmetoz: https://github.com/commercetools/commercetools-adyen-integration/pull/917/commits/9f22b52dfb1d2d595e7a9286781e41c9ebfde520

butenkor (approved):

PR #933, opened 2022-03-04
Update WebComponentsIntegrationGuide.md

BielsasBucket:
Having contacted Adyen support, I found that Klarna could be configured to support auto-capture. You need to request this via a support ticket but have done this and it will work with this extension.


butenkor (approved):
Thanks and sorry for late reply.

PR #943, opened 2022-03-28
Migrate Commonjs files to ES modules.

praveenkumarct:
Migrate all commonjs code to ES modules.

Major changes would be:

1) Change the require statement to import in all the files.
Ex: From const utils = require('./src/utils')
To import utils from './src/utils.js'

2) module exports
Ex:
From module.exports = { processRequest }
To export default { processRequest }

3) Package.json file changes to consider the application files as ES modules
[commercetools-adyen-integration/notification/package.json](https://github.com/commercetools/commercetools-adyen-integration/blob/Es-migration-notification-module/notification/package.json#L5)

4) Read and parse json using fileLoader in [utils](https://github.com/commercetools/commercetools-adyen-integration/pull/943/files#diff-0940ec8e30b28cffe985e6450816c59b057a459fc50bbfde16d4503fb14ad44aR51)

5) Reload module when required(Instead of deleting cache)
https://github.com/commercetools/commercetools-adyen-integration/pull/943/files#diff-af7262c2a89d723433688cdfe6938619058f0d1c84d015d0c467480801c242e9R37


notification/src/config/config.js
```diff
@@ -111,7 +112,7 @@ function loadAndValidateConfig() {
 
 loadAndValidateConfig()
 
-module.exports = {
+export default {
```
lojzatran, line 115: Why do you need default here?

notification/src/config/config.js
```diff
@@ -1,6 +1,7 @@
-const { isEmpty } = require('lodash')
-const loadConfig = require('./config-loader')
+import lodash from 'lodash'
```
lojzatran, line 1:
```suggestion
import { isEmpty } from 'lodash'
```

notification/src/utils/commons.js
```diff
@@ -44,8 +48,18 @@ function getNotificationForTracking(notification) {
...
+async function readAndParseJsonFile(pathToJsonFileFromProjectRoot) {
+  const currentFilePath = fileURLToPath(import.meta.url)
+  const currentDirPath = path.dirname(currentFilePath)
+  const projectRoot = path.resolve(currentDirPath, '../..')
+  const pathToFile = path.resolve(projectRoot, pathToJsonFileFromProjectRoot)
+  const fileContent = await fs.readFile(pathToFile)
+  return JSON.parse(fileContent)
+}
+
+export default {
```
lojzatran, line 60: Why do you have default here?

notification/test/unit/notification.controller.spec.js
```diff
@@ -1,19 +1,16 @@
...
-const logger = require('../../src/utils/logger')
-const ctpClientMock = require('./ctp-client-mock')
-const ctp = require('../../src/utils/ctp')
-const paymentMock = require('../resources/payment-credit-card.json')
-const {
-  buildMockErrorFromConcurrentModificaitonException,
-} = require('../test-utils')
-
+import sinon from 'sinon'
+import lodash from 'lodash'
```
lojzatran, line 2:
```suggestion
import { cloneDeep } from 'lodash'
```

lojzatran (changes requested):

notification/src/config/config.js
praveenkumarct: In few cases where the export needs to be an object, I am using default. (So I am importing config.js as object)
If you have any better way let me know.

notification/src/config/config.js
praveenkumarct:
```
SyntaxError: Named export 'isEmpty' not found. The requested module 'lodash' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'lodash';
const { isEmpty } = pkg;
```
Had to do this way because of above error.

notification/src/utils/commons.js
praveenkumarct: same as above.

notification/test/unit/notification.controller.spec.js
praveenkumarct: same as above.

notification/src/config/config.js
lojzatran: Can you add this comment to the code?

lojzatran (approved):

lojzatran:
And please check the tests.

notification/src/config/config.js
praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/943/commits/9ec1910738df93f3412731231b34890e15733697

Issue #937, opened 2022-03-18
Missing `metadata.ctProjectKey` for some transactions.


wiktor-obrebski:
I found that a small percentage of notifications that our integration got from Adyen does not have `metadata.ctProjectKey` field. 
It breaks the process, as the notification module can not correctly handle such notifications.

In the documentation:
https://github.com/commercetools/commercetools-adyen-integration/blob/master/docs/adr/0003-matching-adyen-notification-in-multitenancy.md

Additionally, I contacted Adyen support. They inform me that sometimes when the notification is sent, they do not have access to transaction meta yet. Therefore, they would need to wait for it, even a few minutes.
In our system, every minute is essential, as we provide services to our customer's houses in 10-30 minutes.

Maybe you can propose something to avoid this issue?
We always use dedicated notification service instances per given Adyen environment. Maybe you can add some optional configuration options, e.g. `defaultAdyenProjectKey` that will automatically assume Adyen project if it is not available in the notification?

lojzatran:
Hi @wiktor-obrebski,

Thank you for contacting us about this problem. 

From what I understood, sometimes with the first notifications they don't contain `metadata.ctProjectKey`, but for the same notifications later on you get with `metadata.ctProjectKey`? And when you test locally, everything is working fine or can you reproduce this problem?

wiktor-obrebski:
It's sporadic. For around ~1 thousand transactions per day, it happens 1-2 times.
There is an example log from notification service (with obscured ids etc.):
```
{..., "level":50,"notification":[{"eventCode":"AUTHORISATION","eventDate":"2022-03-16T22:57:01+01:00","pspReference":"...","success":"true"}],"cause":{"message":"Notification can not be processed as \"metadata.ctProjectKey\"  was not found on the notification.","name":"Error","stack":"Error: Notification can not be processed as \"metadata.ctProjectKey\"  was not found on the notification.\n    at getCtpProjectConfig (/app/src/utils/parser.js:9:11)\n    at handleNotification (/app/src/api/notification/notification.controller.js:24:32)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async Server.<anonymous> (/app/src/server.js:14:9)"},"msg":"Unexpected exception occurred.","time":"2022-03-16T21:57:02.408Z","v":0}
--
```

[...]

PR #952, opened 2022-05-03
Get CTP project key from URL as a fallback

lojzatran:
Workaround for this issue: https://github.com/commercetools/commercetools-adyen-integration/issues/937

Unfortunately I am currently not able to made the fix for AWS Lambda. The problem with AWS is that we don't have access to request in our code and we cannot get the URL: https://github.com/commercetools/commercetools-adyen-integration/blob/master/notification/index.lambda.js. Feel free to suggest sth or we can investigate in the next sprint.

Boy-scout changes:
- I removed the dependency between notification and extension modules in CI. There is no need for them to wait for each other (https://github.com/commercetools/commercetools-adyen-integration/pull/952/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL58)
- I added timeout to fix klarna e2e test: https://github.com/commercetools/commercetools-adyen-integration/pull/952/files#diff-b52e4e13c57ac7ce224efebfc2b86b004c6a569204f7253a9f585253e858cb4eR12

### DIFF
--- a/notification/docs/IntegrationGuide.md
+++ b/notification/docs/IntegrationGuide.md
@@ -27,6 +27,21 @@ Please follow the [instructions](https://docs.adyen.com/development-resources/we
 
 > Note: HMAC verification is enabled by default. You could use "ADYEN_ENABLE_HMAC_SIGNATURE=false" environment variable to disable the verification feature.
 
+### Fallback in case `metadata` is not available
+In rare cases it could happen that notifications do not have `metadata.ctProjectKey` field. Without this field it is not possible to determine which commercetools project the notification belongs to. In order to avoid this rare issue, it is recommended to include the commercetools project key in the path of the URL. In this case, the `public URL` for the Notification module must have the following format:
+```
+https://your-notification-url.com/notifications/${ctp-project-key}
+```
+Notice that the URL path ends with `/notifications/${ctp-project-key}`, where `ctp-project-key` is the project key of the commercetools project that this notification belongs to. This part must always be in the end. See the following examples:
+```
+https://your-notification-url.com/some/other/part/notifications/${ctp-project-key} - valid
+https://your-notification-url.com/notifications/${ctp-project-key}/some/other/part - invalid
+https://your-notification-url.com/notifications/some/other/part/${ctp-project-key} - invalid
+https://your-notification-url.com/notifications/${ctp-project-key}/ - invalid
+```
+
+> Note: if you do not provide `public URL` like above, notification module will still work except for the rare cases.
+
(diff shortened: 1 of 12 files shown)


notification/docs/IntegrationGuide.md
```diff
@@ -27,6 +27,25 @@ Please follow the [instructions](https://docs.adyen.com/development-resources/we
...
+https://your-notification-url.com/notifications/${ctp-project-key}
+```
+
+Notice that the URL path ends with `/notifications/${ctp-project-key}`, where `ctp-project-key` is the project key of the commercetools project that this notification belongs to. This part must always be in the end. See the following examples:
+
+```
+https://your-notification-url.com/some/other/part/notifications/${ctp-project-key} - valid
+https://your-notification-url.com/notifications/${ctp-project-key}/some/other/part - invalid
+https://your-notification-url.com/notifications/some/other/part/${ctp-project-key} - invalid
+https://your-notification-url.com/notifications/${ctp-project-key}/ - invalid
```
leungkinghin, line 44: Where do we verify in our code that 'some/other/part' must be existing in request URL?

leungkinghin (commented):
Looks good to me

notification/docs/IntegrationGuide.md
lojzatran: What do you mean? We do not verify and we don't need to. I explicitly write here that it does NOT work. This is not about if we verify in our code or not. It could be that in some weird use case it will work (which I doubt, but one never knows), but we do not guarantee that. So don't do that. Do it only the way that we considered valid.

leungkinghin (approved):

notification/src/server.js
```diff
@@ -7,8 +7,8 @@ const logger = require('./utils/logger').getLogger()
 function setupServer(routes = defaultRoutes) {
   return http.createServer(async (request, response) => {
     const parts = url.parse(request.url)
-    const route = routes[parts.pathname]
-
+    const path = parts.path?.split('/')?.slice(-2)?.[0] ?? ''
```
LEQADA, line 10: @lojzatran would be nice to have it in a function with a name, which explain what is going on. 🙃

lojzatran: Welcome back @LEQADA, thanks for the comment. I made a change:

https://github.com/commercetools/commercetools-adyen-integration/pull/953/commits/fd8bc8125162136f3665ee33952e3e740c5eae71

LEQADA: So awesome! So fast! Thank you!

PR #960, opened 2022-05-17
Update getting URL path for AWS

lojzatran:
In order to get the URL path for AWS, it is necessary to use `rawPath` attr as documented in the link below.

https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format

So I made this change in this PR.


praveenkumarct (approved):

praveenkumarct:
Please check the build and then merge it. Thanks.

Issue #963, opened 2022-05-25
Cannot run docker image v9.8.0


pawelsadowskiepam:
When starting docker container using version 9.8.0 we have the error
`TypeError: setupServer.listen is not a function
    at file:///app/src/init.js:9:13
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
      at async loadESM (node:internal/process/esm_loader:88:5)
     at async handleMainPromise (node:internal/modules/run_main:61:12)`

There is no such error with version 9.7.0

lojzatran:
Hi @pawelsadowskiepam thank you for reporting the issue. We are checking it and we will let you know once we have more info.

praveenkumarct:
@pawelsadowskiepam new release has been published and v9.8.1 version is available in docker. 

https://github.com/commercetools/commercetools-adyen-integration/releases/tag/v9.8.1

I am closing this issue, Thank you.

PR #964, opened 2022-05-27
Fix syntax issues in the init file.

praveenkumarct:
1) We have an issue with the docker image in the latest release(Since Es module changes are merged into master).
Fix the issue in the init file.
 Issue: https://github.com/commercetools/commercetools-adyen-integration/issues/963

2) Setup-resources in [package.json](https://github.com/commercetools/commercetools-adyen-integration/pull/964/files#diff-0e96fc4c0e4b663545725cc50af120099cb873ea9e3f16eee74d074e39be13feL23) was using require to call setup.js file and that is no more supported. Hence I have added a new library `run-func` which runs the function from terminal instance.  

 Issue : https://github.com/commercetools/commercetools-adyen-integration/issues/965

### DIFF
--- a/extension/package-lock.json
+++ b/extension/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "commercetools-adyen-integration-extension",
-  "version": "9.8.0",
+  "version": "9.8.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


lojzatran (approved):
Don't forget to change the version in package.json.

PR #970, opened 2022-06-06
Make tests running again

lojzatran:
Problem: lots of tests were not running at all after Migration to ES modules. This went completely unnoticed.

https://github.com/mochajs/mocha/issues/2975


praveenkumarct (approved):
Thanks for these changes.

Issue #961, opened 2022-05-20
Indirect Payment Methods (like PayPal) have incorrect CAPTURE status in notification


milindsingh:
We are getting "success": true for a PayPal Capture Failure.

The notification module only validates `"success": true` marking the **Capture** Transaction state in commercetools as `Success`, making it very difficult to identify and monitor.

As we are integrated with PayPal through Adyen Extension we believe the logic to mark Capture as Success or Failure should be done by Notification module.

Notification:

```json
{
    "NotificationRequestItem": {
        "additionalData": {
            "paypalErrorCode": "10606",
            "paypalErrorDescription": "Transaction rejected, please contact the buyer.",
            "metadata.ctProjectKey": "*****",
            "acquirerErrorCode": "10606",
            "paymentMethodVariant": "paypal",
            "acquirerErrorDescription": "Transaction rejected, please contact the buyer.",
            "hmacSignature": "*****"
        },
        "amount": {
            "currency": "AUD",
            "value": 72400
        },
        "eventCode": "CAPTURE",
        "eventDate": "2022-04-13T07:14:36+02:00",
        "merchantAccountCode": "*****",
        "merchantReference": "190000*****-****-*****",
        "originalReference": "*****",
        "paymentMethod": "paypal",
        "pspReference": "*****",
        "success": "true"
    }
}
```

Adyen Support Response:

> Hi Milind,
>  
> When you send us the capture request, our system directly responds with [capture-received]. This means we've received the capture requests from your end. All the modifications (such as captures, refunds, cancellations) are done asynchronously in our system. Our system was built on purpose in this way as authorisation traffic will not be influenced by modification actions. We will send you a CAPTURE notification with success=true once the capture request is processed.
>
> 
>  
> This means that our system expects that the capture will work if:
>  
> 1. The balance is available.
> 2. The transaction is not expired.
> 3. etc.
>
> 
 >
>For Online captures (which are directly submitted to the payment method) as well as Offline captures (which are processed through a batch process) our system uses the same process. If we receive a capture request, and above requirements are met then we submit the CAPTURE notification with success=true.
>
> 
>We will send CAPTURE with success=false in case that these requirements are not met. You will then need to review the reason for the capture failing, fix this, and resubmit the capture request.
 >
>If there would be a 'TECHNICAL' issue with the bank or in this case Paypal, we will return a CAPTURE_FAILED notification at a later >point. Keep in mind that this technically can happen even a couple of days after the capture is submitted.
[...]

PR #971, opened 2022-06-06
#961 enable transaction status change from success to failure

lojzatran:
Fixes: https://github.com/commercetools/commercetools-adyen-integration/issues/961

**Situation**: currently it's not possible for a transaction to go from state `Success` to `Failure`.

**Problem**: in some cases specified in the issue, it's possible that a successful transaction could fail.

**Resolution**: enable changing transaction state from `Success` to `Failure`.


extension/test/integration/get-payment-methods.handler.spec.js
```diff
@@ -5,13 +5,14 @@ import c from '../../src/config/constants.js'
 import config from '../../src/config/config.js'
 import utils from '../../src/utils.js'
 
-describe('::getPaymentMethods::', async () => {
-  const packageJson = await utils.readAndParseJsonFile('package.json')
```
leungkinghin, line 9: What's the actual benefit to put the initialization under `before()`?

docs/adr/006-transaction-update-success-to-failure.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Context
+
+- It was not possible for a payment transaction to change state from `Success` to `Failure`. We thought that once a transaction is successful, it could not fail again.
+- However, in case of `Capture` transaction, the assumption above does not work. This is due to the flow how `Capture` works in Adyen.  
+- Capture has a few stages:
+  - Capture request coming to Adyen.
+  - Capture request coming from Adyen to the bank.
+- Capture notification with `success: true` indicates that the capture request is valid (for example, the authorisation has not expired, and the balance is available) and has been submitted to the bank/third-party processor.
+- Capture notification with `success: false` indicates that the capture request to Adyen has failed. This could be due to a number of reasons for example Insufficient balance on payment.
```
leungkinghin, line 17:
```suggestion
- Capture notification with `success: false` indicates that the capture request to Adyen has failed. This could be due to a number of reasons. For example, Insufficient balance on payment.
```

leungkinghin (approved):
Looks good to me

extension/test/integration/get-payment-methods.handler.spec.js
lojzatran: Benefit is that it didn't work. I fixed it and now it works. So this is the fix. See the explanation here in the Description of this PR: https://github.com/commercetools/commercetools-adyen-integration/pull/970

PR #976, opened 2022-06-16
Replace nyc with c8 tool for code coverage.

praveenkumarct:
- Configure c8 for code coverage and remove nyc from both extension and notification.
- Build is still failing because of the coverage lines restriction in notification module https://github.com/commercetools/commercetools-adyen-integration/pull/976/files#diff-95ef64cb60643f7e9c691d966de57c83e453d4c9097b17cf2e38691bd4220cb0R7

I need your suggestions on the below 2 points:

1) I have moved the [code-coverage check](https://github.com/commercetools/commercetools-adyen-integration/pull/976/files) after unit test run in the build. This I did because the c8 tool is checking only for the latest report (which is e2e tests) and this report has less lines covered.

2) Notification module has only 73.93% code coverage for unit tests. Do we need to reduce the limit or shall we check why the limit is less?


lojzatran:
> * Configure c8 for code coverage and remove nyc from both extension and notification.
> * Build is still failing because of the coverage lines restriction in notification module https://github.com/commercetools/commercetools-adyen-integration/pull/976/files#diff-95ef64cb60643f7e9c691d966de57c83e453d4c9097b17cf2e38691bd4220cb0R7
> 
> I need your suggestions on the below 2 points:
> 
> 1. I have moved the [code-coverage check](https://github.com/commercetools/commercetools-adyen-integration/pull/976/files) after unit test run in the build. This I did because the c8 tool is checking only for the latest report (which is e2e tests) and this report has less lines covered.
> 2. Notification module has only 73.93% code coverage for unit tests. Do we need to reduce the limit or shall we check why the limit is less?

Please check this answer: https://github.com/bcoe/c8/issues/336. My quick idea is this:
1. Add in package.json a test command like following:
```
"test-coverage-run": "npm run test-run && npm run check-coverage",
"test-run": "npm run unit && npm run integration && npm run e2e",
```
2. `npm run test-coverage-run`
3. ....
4. Profit 🎉 

If it works, please modify the names and also try different comands: `npm run cover:unit`. If it doesn't work, try to find sth similar or raise an issue in the repo.

praveenkumarct:
> > * Configure c8 for code coverage and remove nyc from both extension and notification.
> > * Build is still failing because of the coverage lines restriction in notification module https://github.com/commercetools/commercetools-adyen-integration/pull/976/files#diff-95ef64cb60643f7e9c691d966de57c83e453d4c9097b17cf2e38691bd4220cb0R7
> > 
> > I need your suggestions on the below 2 points:
> > 
> > 1. I have moved the [code-coverage check](https://github.com/commercetools/commercetools-adyen-integration/pull/976/files) after unit test run in the build. This I did because the c8 tool is checking only for the latest report (which is e2e tests) and this report has less lines covered.
> > 2. Notification module has only 73.93% code coverage for unit tests. Do we need to reduce the limit or shall we check why the limit is less?
> 
> Please check this answer: [bcoe/c8#336](https://github.com/bcoe/c8/issues/336). My quick idea is this:
> 
> 1. Add in package.json a test command like following:
> 
> ```
> "test-coverage-run": "npm run test-run && npm run check-coverage",
> "test-run": "npm run unit && npm run integration && npm run e2e",
> ```
> 
> 2. `npm run test-coverage-run`
> 3. ....
> 4. Profit 🎉
> 
> If it works, please modify the names and also try different comands: `npm run cover:unit`. If it doesn't work, try to find sth similar or raise an issue in the repo.

Thank you. It works as expected. https://github.com/commercetools/commercetools-adyen-integration/pull/976/commits/cb278c265723e18df4bdd97cf293cf5a4deff415

And these names for the run looks good for me.

extension/package.json
```diff
@@ -5,14 +5,16 @@
...
+    "test-run": "npm run unit && npm run integration && npm run e2e",
+    "check-coverage": "c8 check-coverage --lines 89",
     "test": "npm run check && npm run cover:unit && npm run cover:integration && npm run cover:e2e && npm run check-coverage",
-    "cover:unit": "nyc --silent npm run unit",
-    "cover:integration": "nyc --silent --no-clean npm run integration",
-    "cover:e2e": "nyc --silent --no-clean npm run e2e",
-    "unit": "nyc mocha --file ./test/unit/mock-default-confg.js \"test/unit/**/*.spec.js\"",
-    "integration": "nyc mocha --file ./test/initial-integration-test-set-up.js \"test/integration/**/*.spec.js\"",
-    "e2e": "nyc mocha --file ./test/initial-integration-test-set-up.js \"test/e2e/**/*.spec.js\"",
+    "cover:unit": "npm run unit",
```
lojzatran, line 12: This command is not coverage anymore, right? Seems like redundant now. If so, remove it.

extension/package.json
```diff
@@ -5,14 +5,16 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
-    "check-coverage": "nyc check-coverage --lines 89",
+    "test-coverage-run": "c8 npm run test-run && npm run check-coverage",
+    "test-run": "npm run unit && npm run integration && npm run e2e",
+    "check-coverage": "c8 check-coverage --lines 89",
     "test": "npm run check && npm run cover:unit && npm run cover:integration && npm run cover:e2e && npm run check-coverage",
```
lojzatran, line 11: Is this command used anywhere? If not, remove it.

lojzatran (changes requested):

extension/package.json
praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/976/commits/60686d66e0d61607425d090147573b5566060bc7

extension/package.json
praveenkumarct: I think this will be helpful for local development, So i would prefer leaving it here.
Anyhow in docs also we have mentioned it for development [here](https://github.com/commercetools/commercetools-adyen-integration/blob/master/docs/ContributionGuide.md#development).

praveenkumarct: Makes sense. https://github.com/commercetools/commercetools-adyen-integration/pull/976/commits/bcf6dd8da410d58f83643028c9287fa79b87604d

extension/package.json
```diff
@@ -5,14 +5,13 @@
...
-    "cover:unit": "nyc --silent npm run unit",
-    "cover:integration": "nyc --silent --no-clean npm run integration",
-    "cover:e2e": "nyc --silent --no-clean npm run e2e",
-    "unit": "nyc mocha --file ./test/unit/mock-default-confg.js \"test/unit/**/*.spec.js\"",
-    "integration": "nyc mocha --file ./test/initial-integration-test-set-up.js \"test/integration/**/*.spec.js\"",
-    "e2e": "nyc mocha --file ./test/initial-integration-test-set-up.js \"test/e2e/**/*.spec.js\"",
+    "test-coverage-run": "c8 npm run test-run && npm run check-coverage",
+    "test-run": "npm run unit && npm run integration && npm run e2e",
+    "check-coverage": "c8 check-coverage --lines 89",
+    "test": "npm run check && npm run test-coverage-run",
```
lojzatran, line 11: Now you can remove `test-coverage-run` and put its content to `test` and things will be back as they were before.

extension/package.json
```diff
@@ -5,14 +5,13 @@
...
-    "test": "npm run check && npm run cover:unit && npm run cover:integration && npm run cover:e2e && npm run check-coverage",
-    "cover:unit": "nyc --silent npm run unit",
-    "cover:integration": "nyc --silent --no-clean npm run integration",
-    "cover:e2e": "nyc --silent --no-clean npm run e2e",
-    "unit": "nyc mocha --file ./test/unit/mock-default-confg.js \"test/unit/**/*.spec.js\"",
-    "integration": "nyc mocha --file ./test/initial-integration-test-set-up.js \"test/integration/**/*.spec.js\"",
-    "e2e": "nyc mocha --file ./test/initial-integration-test-set-up.js \"test/e2e/**/*.spec.js\"",
+    "test-coverage-run": "c8 npm run test-run && npm run check-coverage",
+    "test-run": "npm run unit && npm run integration && npm run e2e",
+    "check-coverage": "c8 check-coverage --lines 89",
```
lojzatran, line 10: This command alone doesn't make sense. Please try to remove it and update `test-coverage-run`.

extension/package.json
praveenkumarct: No, It looks good this way. 
test -> which includes check and other test runs with coverage.
test-coverage-run -> without lint check.

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - run: npm ci
       - run: npx lockfile-lint --path package-lock.json --allowed-schemes "https:" "git+https:" --allowed-hosts npm github.com
       - run: npm run check
-      - run: npm run cover:unit
+      - run: npm run test-coverage-run
```
lojzatran, line 28: This change is not the same as before. Before we ran unit test, then we deploy and then we run integration and e2e. Now we run all tests and then deploy. Please check.

extension/package.json
lojzatran: I found it more reasonable to have `cover:unit` etc. for such case than to have a command that cannot be run by itself.

extension/package.json
lojzatran: I don't see a case where you run all the tests and don't want lint check. Even locally it doesn't make sense - when you run all tests, you want to check if it will be ok in CI, and in CI you always have lint check before. So I recommend to remove a command without lint check.

.github/workflows/ci.yml
praveenkumarct: Yes, you are right. I did it on purpose.
Deploying before running tests is not correct I feel.
Why do you need to deploy before running IT and e2e tests?

extension/package.json
praveenkumarct: In CI it is better to have lint check and tests separately. So we have `check` and `test-coverage-run`.
`test` is for local development.

.github/workflows/ci.yml
lojzatran: Because you want to test against your newest changes, not against older version. Probably you don't know, but the tests in CI are running against the deployed serverless deployment in GCP (https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/test/test-utils.js#L25).

So your change means that you will run your code against old version in GCP and then deploy the changes after the tests to GCP.

extension/package.json
lojzatran: I don't see what is the benefit of having 2 commands if you never use them separately. WDYT?

.github/workflows/ci.yml
lojzatran: I'm not sure if I understand you correctly. Can't you just move this line after deployment is done to line 51?

praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/976/commits/f3a9bf03b3ff79af16a3d3976dce3a0a64886fe8

extension/package.json
praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/976/commits/2057cb475054f2c07c04dd50eae390c28e6e70a4

extension/package.json
praveenkumarct: cover:unit is a different thing, but keeping this command `check-coverage` is also for refactoring to make it more clear for reading. 
So I would keep it this way.

lojzatran (approved):

PR #992, opened 2022-07-24
Update node fetch

praveenkumarct:
- Upgrade Node-Fetch library to latest version.
- Fix tests which were using stub, Use nock instead.


lojzatran (commented):
Please check failing tests

leungkinghin (approved):
Looks good to me
Please make sure test is ok after localtunnel resumed

PR #998, opened 2022-08-08
Improve error logging

lojzatran:
I added name of the module (extension) to make the error clear that it comes from the extension module.

Boy-scout change: I also put back the dependency between build of extension and notification modules. Now the notification module will wait for build from the extension module to complete before it starts. To be frank the 2 modules do not depend on each other, but I see that the builds are failing randomly, so maybe that could be a problem. Let's see if it helps.

### DIFF
--- a/extension/src/utils.js
+++ b/extension/src/utils.js
@@ -38,7 +38,9 @@ function getLogger() {
 }
 
 function handleUnexpectedPaymentError(paymentObj, err) {
-  const errorMessage = `Unexpected error (Payment ID: ${paymentObj?.id}): ${err.message}.`
+  const errorMessage =
+    '[commercetools-adyen-integration-extension] ' +
+    `Unexpected error (Payment ID: ${paymentObj?.id}): ${err.message}.`
   const errorStackTrace = `Unexpected error (Payment ID: ${
     paymentObj?.id
   }): ${JSON.stringify(serializeError(err))}`
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
   notification_build:
     name: Notification Module
     concurrency: notification_build
+    needs: extension_build
     runs-on: ubuntu-latest
     env:
       ADYEN_INTEGRATION_CONFIG: ${{ secrets.ADYEN_INTEGRATION_CONFIG }}
--- a/extension/test/unit/payment.controller.spec.js
+++ b/extension/test/unit/payment.controller.spec.js
@@ -134,6 +134,10 @@ describe('Payment controller', () => {
       expect(data.errors).to.not.empty
       expect(data.errors).to.have.lengthOf(1)
       expect(data.errors[0].code).to.equal('General')
+      expect(data.errors[0].message).to.equal(
+        '[commercetools-adyen-integration-extension] ' +
+          'Unexpected error (Payment ID: undefined): request.on is not a function.'
+      )
     }
 
     sinon.stub(utils, 'sendResponse').callsFake(sendResponse)


leungkinghin (approved):
It looks good to me

extension/src/utils.js
```diff
@@ -38,7 +38,9 @@ function getLogger() {
 }
 
 function handleUnexpectedPaymentError(paymentObj, err) {
-  const errorMessage = `Unexpected error (Payment ID: ${paymentObj?.id}): ${err.message}.`
+  const errorMessage =
+    '[commercetools-adyen-integration-extension] ' +
```
praveenkumarct, line 42: I think this is not the resolution of this ticket. We always had a separate builds for Extension and Notification modules, So the error was very clear to know from which module it is coming from.

May be we wanted to have a clear error message in the logs, what exactly in the config is missing.

`BadRequest: Unexpected error (Payment ID: d84aa096-cbb4-4379-96b8-e670c1b5fb3f): Configuration is not provided. Please update the configuration. ctpProjectKey: [adyen-qa-2].` 

In the above error, Configuration is provided, but it has some configs missing.

praveenkumarct (changes requested):

extension/src/utils.js
lojzatran: @praveenkumarct AFAIK this problem was triggered by you last sprint because you had a problem with failing tests and you saw a message that config was suddenly missing. So you updated it in github secrets, right? Did it work after you update the config? What was exactly the problem?

The message exactly says what config is missing. A config for `ctpProjectKey: [adyen-qa-2]`.

When reading the JIRA task, you could see a complication like this: `When the test are run by the GitHub actions we do not know where the config is missing: function or the test itself`. So it's not about if config is missing an attribute, but rather we didn't know if the missing config error above comes from the extension function or the test.

praveenkumarct: > The message exactly says what config is missing. A config for ctpProjectKey: [adyen-qa-2].
I couldn't understand the error from the logs. Since I had already provided [adyen-qa-2] config properly.

So now the error looks like below:

`[commercetools-adyen-integration-extension] BadRequest: Unexpected error (Payment ID: d84aa096-cbb4-4379-96b8-e670c1b5fb3f): Configuration is not provided. Please update the configuration. ctpProjectKey: [adyen-qa-2].`

For all the developers invovled now in this topic, will understand the issue next time. But If any new person checks this error, may be it is confusing for him again.

Can we add a point in Readme or somewhere else, Informing the users or devs to ensure they are using the same configs both in tests and Extension function?
> I am still okay if you don't want this.

lojzatran: > > The message exactly says what config is missing. A config for ctpProjectKey: [adyen-qa-2].
> > I couldn't understand the error from the logs. Since I had already provided [adyen-qa-2] config properly.
> 
> So now the error looks like below:
> 
> `[commercetools-adyen-integration-extension] BadRequest: Unexpected error (Payment ID: d84aa096-cbb4-4379-96b8-e670c1b5fb3f): Configuration is not provided. Please update the configuration. ctpProjectKey: [adyen-qa-2].`
> 
> For all the developers invovled now in this topic, will understand the issue next time. But If any new person checks this error, may be it is confusing for him again.
> 
> Can we add a point in Readme or somewhere else, Informing the users or devs to ensure they are using the same configs both in tests and Extension function?
> 
> > I am still okay if you don't want this.

We do not have a doc for developers in such details. Where would you add it and what should be there? Also, it's a different task, so I could create a ticket for it if you see a need.

Also what you said here: `Can we add a point in Readme or somewhere else, Informing the users or devs to ensure they are using the same configs both in tests and Extension function?` - this is true only for some tests because you are making request to the extension for a particular project, so adyen extension needs to have this project set up. If you want to make it fail and test that it fails, then your statement would not apply. So it depends on what you test :)

praveenkumarct: Anyhow, Since this is from the tests. You are right. We can close this topic. Thank you for the explanation :)

praveenkumarct (approved):

PR #1010, opened 2022-09-03
Generate idempotency key

lojzatran:
This PR adds automatic idempotency key generation.

Note: tests are currently failing. The problem is I am testing generating idempotency key and in order to activate this feature, I need to set a config `generateIdempotencyKey`. This config I can set locally just by overriding the config from the integration test. However, this is not possible in github as we are using GCP functions and the test is running on a different machine than the extension deployment. This problem will be fixed once I update the config.

**Do NOT merge this PR until tests are fixed ⚠️**


extension/docs/Refund.md
```diff
@@ -181,6 +181,12 @@ Follow these recommendations when using the idempotency key:
 - Generate idempotency keys using the version 4 (random) UUID type to prevent two API credentials under the same account from accessing each others responses.
```
leungkinghin, line 181:
```suggestion
- Generate idempotency keys using the version 4 (random) UUID type to prevent two API credentials under the same account from accessing each other's responses.
```

extension/docs/Refund.md
```diff
@@ -181,6 +181,12 @@ Follow these recommendations when using the idempotency key:
...
 - Use [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) when retrying.
 
+#### Generating idempotency key by adyen-integration
+
+adyen-integration can automatically generate unique idempotency key for refund requests. It uses transaction ID as the idempotency key.
+
+In order to use this feature, set the option [`generateIdempotencyKey=true`](./HowToRun.md#optional-attributes). For every refund request adyen-integration will add an idempotency key if the custom field `idempotencyKey` is not present. If the custom field `idempotencyKey` is present, its value will be taken as the idempotency key.
+
```
leungkinghin, line 189: I think you can include following Adyen documentation as same as `ManualCapture.md`

> For in-depth information about API idempotency in Adyen, see [Adyen's documentation](https://docs.adyen.com/development-resources/api-idempotency).

extension/src/paymentHandler/payment-utils.js
```diff
@@ -190,6 +190,13 @@ function isValidMetadata(str) {
   return str.indexOf(' ') < 0
 }
 
+function getIdempotencyKey(transaction) {
+  let idempotencyKey = transaction.custom?.fields?.idempotencyKey
```
leungkinghin, line 194: where do you initialize `transaction.custom.fields.idempotencyKey`? I cannot find it in you PR

leungkinghin (changes requested):

extension/docs/Refund.md
lojzatran: I had it here like you suggested but I moved it to the bottom. The reason is when the link is here it looks like the user has to check the docs to use the Refund. That is not true and the documentation is rather a source where we got the info. User does not have to care about it at all. Therefore I moved to the bottom so it's not confusing.

leungkinghin: ok i found it now. It is not in this pr so i didn't realize you added it.

extension/src/paymentHandler/payment-utils.js
leungkinghin: I saw [this part](https://github.com/commercetools/commercetools-adyen-integration/blob/d349060ef9a8eafac34429f981904d338b5b95bf/extension/src/service/web-component-service.js#L65) and it is clear for me now. I thought the program would put the generated `idempotencyKey` back to transasction custom field at some points

leungkinghin (approved):

extension/src/paymentHandler/payment-utils.js
lojzatran: I see. Let me check if this could be confusing for the user or not.

lojzatran: @leungkinghin https://github.com/commercetools/commercetools-adyen-integration/pull/1010/commits/1aadf8b15b7f2a8e3c07de939600a4fa275995af

PR #1014, opened 2022-09-14
Conditional extension trigger

lojzatran:
In this PR I added [Conditional trigger for API Extension](https://docs.commercetools.com/api/projects/api-extensions#conditional-triggers).

Apart from the regular tests, I tested the feature by creating a payments without `paymentMethodInfo`
```json
{
  "key": "testPayment1-1663184584232",
  "amountPlanned": {
    "type": "centPrecision",
    "currencyCode": "USD",
    "centAmount": 1000,
    "fractionDigits": 2
  }
}
```
and with different `paymentInterface`
```json
{
  "key": "testPayment1-1663184584232",
  "amountPlanned": {
    "type": "centPrecision",
    "currencyCode": "USD",
    "centAmount": 1000,
    "fractionDigits": 2
  },
  "paymentMethodInfo": {
    "paymentInterface": "pay1",
    "method": "affirm",
    "name": {
      "en": "Affirm"
    }
  }
}
```

Both was working.

Question: I also removed the code that check the paymentMethodInfo. Is it a breaking change for the customer? Now they need to use the conditional trigger, or the extension module won't work correctly. WDYT?


extension/resources/api-extension.json
```diff
@@ -7,7 +7,8 @@
   "triggers": [
     {
       "resourceTypeId": "payment",
-      "actions": ["Create", "Update"]
+      "actions": ["Create", "Update"],
+      "condition": "paymentMethodInfo is defined AND paymentMethodInfo(paymentInterface is defined) AND paymentMethodInfo(paymentInterface=\"ctp-adyen-integration\")"
```
leungkinghin, line 11: if `paymentMethodInfo(paymentInterface=\"ctp-adyen-integration\")` is true, then `paymentMethodInfo(paymentInterface is defined)` should be true, so we don't need the middle condition checking, isn't it?

leungkinghin (changes requested):
I agree it is a breaking change too, as long as customer has to update the extension resource in their CTP project. Otherwise the `paymentMethodInfo` and `paymentInterface` checking would be no longer there.

leungkinghin (approved):

extension/resources/api-extension.json
lojzatran: It would be too easy if it works like you suggested. Check here in the docs the orange box: https://docs.commercetools.com/api/projects/api-extensions#conditional-triggers

TBH I also tried your suggestion at first and it threw an error 😬 that's how I came to this condition in the end.

butenkor:
Just for the sake of compatibility I could leave the check since it is just one liner, wdyt?

lojzatran:
> Just for the sake of compatibility I could leave the check since it is just one liner, wdyt?

I removed it already and released it as a breaking change. I was thinking to leave it there, but then there is a high chance it would stay there forever. So I decided to remove it now.

extension/src/config/init/ensure-api-extensions.js
```diff
@@ -40,19 +40,13 @@ async function ensureApiExtensions(
         'Successfully created an API extension for payment resource type ' +
           `(key=${apiExtensionTemplate.key})`
       )
-    } else if (
-      !_.isEqual(existingExtension.destination, extensionDraft.destination)
-    ) {
+    } else {
+      const actions = buildUpdateActions(existingExtension, extensionDraft)
```
butenkor, line 44: What if actions array is empty? Should not we check it before trying to update?

lojzatran: https://github.com/commercetools/commercetools-adyen-integration/pull/1017/files

PR #1037, opened 2022-11-14
Notification management api

lojzatran:
This PR adds an ensure functionality for adyen webhooks.

HMAC is not added yet in this PR. It will be done in the next PR.

The part `ensuring fallback for metadata` is not possible with this feature. The reason is there is no specified connection between the Adyen project and CTP project and thus the URL cannot be ensured automatically. The correct fallback URL must be created by the user.

### DIFF
--- a/notification/docs/HowToRun.md
+++ b/notification/docs/HowToRun.md
@@ -99,13 +99,15 @@ Other configurations can be set as direct child attributes in `ADYEN_INTEGRATION
 }
 ```
 
-| Name                         | Content                                                                                                                                                                         | Required | Default value                                                                                         |
-| ---------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | ----------------------------------------------------------------------------------------------------- |
-| `adyenPaymentMethodsToNames` | Key-value object where key is `paymentMethod` returned in the notification and value is the custom localized name that will be saved in CTP `payment.paymentMethodInfo.method`. | NO       | `{scheme: {en: 'Credit Card'}, pp: {en: 'PayPal'}, klarna: {en: 'Klarna'}, gpay: {en: 'Google Pay'}}` |
-| `port`                       | The port number on which the application will run.                                                                                                                              | NO       | 443                                                                                                   |
-| `logLevel`                   | The log level (`trace`, `debug`, `info`, `warn`, `error`, `fatal`).                                                                                                             | NO       | `info`                                                                                                |
-| `keepAliveTimeout`           | Milliseconds to keep a socket alive after the last response ([Node.js docs](https://nodejs.org/dist/latest/docs/api/http.html#http_server_keepalivetimeout)).                   | NO       | Node.js default (5 seconds)                                                                           |
-| `removeSensitiveData`        | Boolean attribute. When set to "false", Adyen fields with additional information about the payment will be saved in the interface interaction and in the custom fields.         | NO       | true                                                                                                  |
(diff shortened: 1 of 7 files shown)


notification/docs/HowToRun.md
```diff
@@ -99,13 +99,15 @@ Other configurations can be set as direct child attributes in `ADYEN_INTEGRATION
...
+| `removeSensitiveData`        | Boolean attribute. When set to "false", Adyen fields with additional information about the payment will be saved in the interface interaction and in the custom fields.                                      | true                                                                                                  |
+| `notificationBaseUrl`        | Publicly available URL of the Notification module. In case of any payment changes, Adyen will call this URL and pass the notification in body. This attribute is used when calling `npm run setup-resources` |                                                                                                       |
+| `apiKey`                     | This key will be used to make authenticated API requests to Adyen. This attribute is used when calling `npm run setup-resources`                                                                             |                                                                                                       |
```
praveenkumarct, line 110: Description just says where and how this `apiKey` is used, but It would be good if it explains what it is.

notification/src/config/init/ensure-adyen-webhook.js
```diff
@@ -0,0 +1,99 @@
...
+      `https://management-test.adyen.com/v1/merchants/${merchantId}/webhooks`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Api-Key': adyenApiKey,
+        },
+      }
+    )
+    const getWebhookResponseResponseJson = await getWebhookResponse.json()
```
praveenkumarct, line 32:
```suggestion
    const getWebhookResponseJson = await getWebhookResponse.json()
```
It looks better this way.

notification/src/config/init/ensure-adyen-webhook.js
```diff
@@ -0,0 +1,99 @@
+import fetch from 'node-fetch'
+import { serializeError } from 'serialize-error'
+import { getLogger } from '../../utils/logger.js'
+import config from '../config.js'
+
+const mainLogger = getLogger()
+
+async function ensureAdyenWebhook(adyenApiKey, webhookUrl, merchantId) {
```
praveenkumarct, line 8: Can we write some tests to make sure this code doesn't fail?

praveenkumarct (changes requested):

notification/docs/HowToRun.md
lojzatran: OK, what would you write here?

notification/src/config/init/ensure-adyen-webhook.js
lojzatran: I could write the tests, but it would take time and the benefit is almost zero. This code is rarely changed and it just ensures one resource, so it's very straightforward to test manually.

notification/docs/HowToRun.md
praveenkumarct: I am not sure what is `apiKey` here. 
Ex: for 'notificationBaseUrl' we wrote,  `Publicly available URL of the Notification module`.
Same way a short sentence which explains about this key(where can someone get or generate this from).

lojzatran: It is actually `A String that is used to make authenticated API requests to Adyen` which is more or less what is there.

praveenkumarct: Okay anyhow i think it is also okay to leave it like this then.

praveenkumarct (approved):

PR #1040, opened 2022-11-18
Automatically generate hmac key

lojzatran:


notification/docs/IntegrationGuide.md
```diff
@@ -23,7 +23,7 @@ The following diagram shows the integration flow of the notification module base
...
 
-You can set up the webhooks and generate HMAC by running the command `npm run setup-resources` as below, the command requires the `ADYEN_INTEGRATION_CONFIG` to be set as an environment variable. Be aware that this command also sets up [the commercetools resources required for the notification module](./HowToRun.md#commercetools-project-requirements).
+You can set up the webhooks and generate HMAC by running the command `npm run setup-resources` as below, the command requires the `ADYEN_INTEGRATION_CONFIG` to be set as an environment variable. For every project you want to generate HMAC key set the attribute `enableHmacSignature: true` and do not have the attribute `secretHmacKey`. The new HMAC key will be generated and the updated `ADYEN_INTEGRATION_CONFIG` will be printed to you. Use this updated `ADYEN_INTEGRATION_CONFIG` to run the notification module. Be aware that this command also sets up [the commercetools resources required for the notification module](./HowToRun.md#commercetools-project-requirements).
```
leungkinghin, line 26: Does it mean we need to update the environment variable after we have received the updated `ADYEN_INTEGRATION_CONFIG`?

notification/src/config/init/ensure-adyen-webhook.js
```diff
@@ -59,41 +60,57 @@ async function ensureAdyenWebhook(adyenApiKey, webhookUrl, merchantId) {
...
+          adyenMerchantId,
+          webhookId
+        )
+        jsonConfig.adyen[adyenMerchantId].secretHmacKey = hmacKey
+      }
     }
   }
+
+  mainLogger.info('Set the following environmental variable')
+  console.log(`ADYEN_INTEGRATION_CONFIG='${JSON.stringify(jsonConfig)}'`)
```
leungkinghin, line 113: is it possible to update the environment variable for user here instead of printing it out?

notification/docs/IntegrationGuide.md
lojzatran: `Use this updated `ADYEN_INTEGRATION_CONFIG` to run the notification module.` - is this sentence clear? If not, please suggest how should I change it.

notification/src/config/init/ensure-adyen-webhook.js
lojzatran: Not sure. Can you make a suggestion how to do that?

leungkinghin: > Not sure. Can you make a suggestion how to do that?

e.g. `process.env.ADYEN_INTEGRATION_CONFIG = jsonConfig`

leungkinghin: > 

In case we need to let user know the generated HMAC key, we can update ADYEN_INTEGRATION_CONFIG in environment variable first and then print it out. wdyt?

leungkinghin: > One thing to be aware of: HMAC key is kind of CTP client secret - once it is generated, you cannot get the same key anymore if you refresh the page. Same idea with HMAC key - you can only generate new HMAC keys, not get the existing one from Adyen.

If we need to let user know the generated HMAC key since it cannot be retrieved later, we can update `ADYEN_INTEGRATION_CONFIG` in environment variable first and then print it out. wdyt?

notification/docs/IntegrationGuide.md
leungkinghin:
```suggestion
You can set up the webhooks and generate HMAC by running the command `npm run setup-resources` as below, the command requires the `ADYEN_INTEGRATION_CONFIG` to be set as an environment variable. For every project you want to generate HMAC key set the attribute `enableHmacSignature: true` and do not have the attribute `secretHmacKey`. The new HMAC key will be generated and the updated `ADYEN_INTEGRATION_CONFIG` will be printed to you. Use this updated `ADYEN_INTEGRATION_CONFIG` to replace the existing one in environment variable and run the notification module. Be aware that this command also sets up [the commercetools resources required for the notification module](./HowToRun.md#commercetools-project-requirements).
```

notification/src/config/init/ensure-adyen-webhook.js
lojzatran: This means I have to run this HMAC generation before the start the notification module, right? Because right now this update runs separately and manually by the user. And how do you ensure that you always have the same HMAC key even after restart?

leungkinghin (approved):

PR #1043, opened 2022-12-01
Set up webhooks for tests

lojzatran:
# How the set up is done
- For local I start the notification server and localtunnel and ensure webhook in Adyen.

- For CI I use the existing GCP Function for notification.

# Challenges in webhook tests

## It is not possible to enable/disable HMAC key verification on the fly for notification module in CI. - NOT DONE
As you know, we have HMAC key verification for notifications coming from Adyen. I tried to disable HMAC key verification for notification module in CI to successfully run the tests. It is not possible to disable or enable HMAC key per test as tests are running on github and notification is deployed in GCP. Because I disabled HMAC, one test in notification is failing that is testing HMAC. So far I could not find an easy solution for this issue. The only way I could think of is to deploy a separate instances of notification module for webhook tests.

## It is necessary to remove webhook in Adyen after tests
If webhook is not removed, it would receive notifications (e.g. when someone is testing manually or run tests locally/on CI) and try to call non-existing URL. In this case, webhook would get stuck after some time and won't be able to run anymore unless someone manually cleans it. Thus it needs to be removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
       ADYEN_INTEGRATION_CONFIG: ${{ secrets.ADYEN_INTEGRATION_CONFIG }}
       CI_EXTENSION_BASE_URL: ${{ secrets.CI_EXTENSION_BASE_URL }}
       CI_NOTIFICATION_URL: ${{ secrets.CI_NOTIFICATION_URL }}
-      ENABLE_HMAC_SIGNATURE: false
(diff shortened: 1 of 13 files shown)


praveenkumarct (approved):
LGTM !!

merijndebeer-adyen:
Thank you for your contribution. We have reviewed this pull request and decided to not merge it since it does not fit in the current version of the integration. Regards Adyen Product Management

PR #1050, opened 2022-12-22
Add create session handler and remove /paymentMethods, /payments, and /payments/details handlers

praveenkumarct:
- Add create sessionRequest handler and unit tests.
- Remove `/paymentMethods`(**get-payment-methods.handler.js**) and their unit test and Integration test.


extension/src/paymentHandler/make-lineitems-payment.handler.js
```diff
@@ -1,133 +0,0 @@
...
-  )
-  const commercetoolsProjectKey =
-    paymentObject.custom.fields.commercetoolsProjectKey
-  if (!makePaymentRequestObj.lineItems) {
-    const ctpCart = await _fetchMatchingCart(
-      paymentObject,
-      commercetoolsProjectKey
-    )
-    if (ctpCart) {
-      makePaymentRequestObj.lineItems = createLineItems(paymentObject, ctpCart)
```
butenkor, line 22: Careful, since we are removing an existing convenience features like automatic lineItem generation out of the cart data here! I would probably reuse both files that you removed and did some adjustments/renaming since majority of the logic stays the same. Sessions and just different name for makePymentRequest/response.

extension/src/paymentHandler/payment-handler.js
```diff
@@ -73,33 +62,18 @@ function _getPaymentHandlers(paymentObject) {
...
   )
     handlers.push(getCarbonOffsetCostsHandler)
-  if (
-    paymentObject.custom.fields.makePaymentRequest &&
-    !paymentObject.custom.fields.makePaymentResponse
-  ) {
-    const makePaymentRequestObj = JSON.parse(
-      paymentObject.custom.fields.makePaymentRequest
-    )
-    if (_requiresLineItems(makePaymentRequestObj))
```
butenkor, line 93: line items support is still required.

butenkor (changes requested):

extension/src/paymentHandler/sessions-request.handler.js
```diff
@@ -3,26 +3,29 @@ import {
   createSetCustomFieldAction,
 } from './payment-utils.js'
 import c from '../config/constants.js'
-import { getPaymentMethods } from '../service/web-component-service.js'
+import { createSessionRequest } from '../service/web-component-service.js'
 
 async function execute(paymentObject) {
-  const getPaymentMethodsRequestObj = JSON.parse(
-    paymentObject.custom.fields.getPaymentMethodsRequest
+  const amountUpdatesRequestObj = JSON.parse(
```
leungkinghin, line 9: Why is it named as `amountUpdatesRequestObj` ?

extension/test/unit/create-session-request.handler.spec.js
```diff
@@ -45,43 +45,45 @@ describe('get-payment-methods::execute::', () => {
   })
 
   it('handlePayment should return the right actions', async () => {
-    const adyenGetPaymentResponse = {
+    const adyenGetSessionResponse = {
       paymentMethods: [
```
leungkinghin, line 49: I think the `paymentMethods` attribute is not included in response from /session endpoint.
https://docs.adyen.com/api-explorer/Checkout/latest/post/sessions

Perhaps you can make a CURL command to this adyen endpoint and verify it?

extension/test/unit/payment-handler.spec.js
```diff
@@ -209,7 +38,7 @@ describe('payment-handler::execute', () => {
 
   describe('amountPlanned', () => {
     it(
-      'is different than the amount in makePayment request custom field and interface interaction is empty, ' +
+      'is different than the amount in create session request custom field and interface interaction is empty, ' +
```
leungkinghin, line 41: Do we need to change the implementation of this test case? I found it is still accessing  `ctpPaymentClone.custom.fields.makePaymentRequest`

leungkinghin (changes requested):
1. Line item feature is still required and we need to check how to add it into new flow.
2. We still need `multitenancy.spec.js` and `payment-handler-authorization.spec.js` as it is not about web component v5 migration.

extension/test/unit/create-session-request.handler.spec.js
praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/1050/commits/8312d730202018c80d967ac0b4c25dd7adbfe75f

Yes, you are right.

extension/src/paymentHandler/sessions-request.handler.js
praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/1050/commits/0d6fb80dacde447429ffa26a8514864374db90ce#diff-a6ebee5cac76979c3469c418f46614bf4d3f8de5a426108afd27c1b292e55a80

extension/test/unit/payment-handler.spec.js
praveenkumarct: https://github.com/commercetools/commercetools-adyen-integration/pull/1050/commits/ee56b5d6d6d7cd8a46d4cdd9a801f972df458681

leungkinghin: Aren't we going to use create session to test the `amountPlanned` here? I think we don't use `makePayment` thingy anymore in this test case.

praveenkumarct: Follow up ticket created to handle removal of makePayment request handler.
https://jira.commercetools.com/browse/CTPI-761

And this test is specific to makePayment.

extension/src/paymentHandler/make-lineitems-payment.handler.js
praveenkumarct: Added back these changes and since they are complicated to remove, I have created a follow up ticket to discuss with the team and migrate later.
https://jira.commercetools.com/browse/CTPI-761

extension/src/paymentHandler/payment-handler.js
praveenkumarct: https://jira.commercetools.com/browse/CTPI-761

leungkinghin: @praveenkumarct If we follow up this part as next step, I would suggest to keep this part this moment, so that it helps us to understand how to migrate line-item features for web component v5.0

In addition, please keep `multitenancy.spec.js` and `payment-handler-authorization.spec.js`. They are required all the time no matter which web component version we are using.

extension/src/paymentHandler/make-lineitems-payment.handler.js
leungkinghin: @praveenkumarct If we follow up this part as next step, I would suggest to keep this handler this moment, so that it helps us to understand how to migrate line-item features for web component v5.0

In addition, please keep `multitenancy.spec.js` and `payment-handler-authorization.spec.js`. They are required all the time no matter which web component version we are using.

praveenkumarct: Yes, I have already reverted these changes back.

extension/src/paymentHandler/payment-handler.js
praveenkumarct: Same here.

leungkinghin (approved):

PR #1051, opened 2023-01-02
Add Integration tests for session creation

praveenkumarct:
- Remove Affirm and Klarna make payment request Integration tests.
- Add Integration test for create session handler.


leungkinghin:
Also, we may need to add transaction into payment after creating payment session. I am not sure since we don't have `pspReference` in extension. Let's discuss with the team about this point.

https://github.com/commercetools/commercetools-adyen-integration/blob/dbd47fb9473c93028235fe7ba72e9327c3c541bf/extension/src/paymentHandler/make-payment.handler.js#L54

extension/test/integration/sessions-request.handler.spec.js
```diff
@@ -0,0 +1,111 @@
...
+        },
+      }
+      expect(statusCode).to.equal(201)
+
+      const { createSessionRequest, createSessionResponse } =
+        payment.custom.fields
+      expect(createSessionRequest).to.be.deep.equal(
+        JSON.stringify(createSessionRequestDraft)
+      )
+
```
leungkinghin, line 81: Please add assertion to make sure the payment key is equal to the reference inside create session request

leungkinghin (changes requested):

extension/test/integration/sessions-request.handler.spec.js
praveenkumarct: [52b0e90](https://github.com/commercetools/commercetools-adyen-integration/pull/1051/commits/52b0e9088c1b3c7c856d7d5a0d768bb42814fd97)

praveenkumarct:
>Also, we may need to add transaction into payment after creating payment session. I am not sure since we don't have pspReference in extension. Let's discuss with the team about this point.

As discussed with the team, We will not add transaction for CreateSessionRequest.
CreateSessionResponse doesn't contains 'resultCode' field.

leungkinghin (approved):

Issue #1052, opened 2023-01-03
Refunding Request Obj is sometimes wrong


BielsasBucket:
When performing refunds (our DEV examples below) we are seeing that sometimes these fail and produce the following requests (taken from the interface interactions):

```
{
            "type": {
                "typeId": "type",
                "id": "4c19a616-a212-4d14-bc0c-4b569c3e36fc"
            },
            "fields": {
                "createdAt": "2022-12-28T22:15:59.930Z",
                "response": "{\"name\":\"SyntaxError\",\"message\":\"Unexpected token M in JSON at position 4\",\"stack\":\"SyntaxError: Unexpected token M in JSON at position 4\\n    at JSON.parse (<anonymous>)\\n    at Response.json (file:///app/node_modules/node-fetch/src/body.js:149:15)\\n    at runMicrotasks (<anonymous>)\\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\\n    at async fetchAsync (file:///app/src/service/web-component-service.js:216:24)\\n    at async callAdyen (file:///app/src/service/web-component-service.js:182:35)\\n    at async file:///app/src/paymentHandler/refund-payment.handler.js:37:37\\n    at async Promise.all (index 0)\\n    at async Object.execute (file:///app/src/paymentHandler/refund-payment.handler.js:24:3)\\n    at async Promise.all (index 0)\"}",
                "request": "{\"body\":\"{\\\"modificationAmount\\\":{\\\"value\\\":1,\\\"currency\\\":\\\"GBP\\\"},\\\"originalReference\\\":\\\"T4ZNQZZTCV5X8N82\\\",\\\"reference\\\":\\\"a810873b-e394-45d1-9a61-c4a2a24b0d12\\\",\\\"metadata\\\":{\\\"ctProjectKey\\\":\\\"fa-compcomm-dev\\\"},\\\"merchantAccount\\\":\\\"S0115_Kings-Will-Dream-Web\\\"}\"}",
                "type": "refund"
            }
        }
```

Most of the time this works resulting in an interface interaction such as:

```
{
            "type": {
                "typeId": "type",
                "id": "4c19a616-a212-4d14-bc0c-4b569c3e36fc"
            },
            "fields": {
                "createdAt": "2022-12-28T22:15:42.032Z",
                "response": "{\"pspReference\":\"NDJBCR5QC3RZNN82\",\"response\":\"[refund-received]\"}",
[...]

PR #1059, opened 2023-01-18
Handle non-JSON response

leungkinghin:
Fixes : #1052 

- When we receive the response from Adyen API call, now we check if the response is JSON format. If it isn't, we would put it as plain text response into `InterfaceInteractions` so that user would know what actually happen.

- Fix other refund unit test cases since mocking response for payment API call is not applicable.

### DIFF
--- a/extension/src/service/web-component-service.js
+++ b/extension/src/service/web-component-service.js
@@ -225,7 +225,9 @@ async function fetchAsync(
   } catch (err) {
     if (response)
       // Handle non-JSON format response
-      responseBody = responseBodyInText
+      throw new Error(
+        `Unable to receive non-JSON format resposne from Adyen API : ${responseBodyInText}`
+      )
     // Error in fetching URL
     else throw err
   } finally {
--- a/extension/src/service/web-component-service.js
+++ b/extension/src/service/web-component-service.js
@@ -192,6 +192,7 @@ async function callAdyen(
     returnedRequest = { body: JSON.stringify(requestArg) }
     returnedResponse = serializeError(err)
   }
+
   return { request: returnedRequest, response: returnedResponse }
 }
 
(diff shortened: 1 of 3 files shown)


extension/src/service/web-component-service.js
```diff
@@ -206,18 +207,33 @@ async function fetchAsync(
...
+
+  try {
+    response = await fetch(url, request)
+    responseBodyInText = await response.text()
+
+    responseBody = JSON.parse(responseBodyInText)
+  } catch (err) {
+    if (response)
+      // Handle non-JSON format response
+      responseBody = responseBodyInText
```
butenkor, line 228: If we do not throw an error here it will be processed as if the Adyen response was successful? Can the caller of this function work with the `responseBodyInText` which it assumed previously to be a valid JSON?

leungkinghin: Even error is thrown here, the error stack will be serialized as String and return as response in following line
https://github.com/commercetools/commercetools-adyen-integration/blob/e6bf765031d337e87f02bfb3a6ce824dd552ca78/extension/src/service/web-component-service.js#L193

The error stack then is saved into response field in `InterfaceInteraction`. It seems that no difference from without throwing error.

leungkinghin: I changed to throwing error, as the error stack contains line number for investigation. It would be more useful than simply returning the non-JSON response
https://github.com/commercetools/commercetools-adyen-integration/pull/1059/commits/31d259f17b1e65b5179755c8e407ea666ccd74ec

extension/test/unit/refund-payment.handler.spec.js
```diff
@@ -44,16 +43,14 @@ describe('refund-payment::execute', () => {
 
   beforeEach(() => {
     const adyenConfig = config.getAdyenConfig(adyenMerchantAccount)
-    scope = nock(`${adyenConfig.apiBaseUrl}`)
+    scope = nock(`${adyenConfig.legacyApiBaseUrl}/Payment/v64`)
```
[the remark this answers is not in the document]
leungkinghin: https://github.com/commercetools/commercetools-adyen-integration/pull/1059/commits/0c4021cd08c52810fe795871d28bbfd9aebedbb4

butenkor (approved):

praveenkumarct (approved):

PR #1072, opened 2023-02-06
V5 migration klarna capture payment

leungkinghin:
- Fix capture payment flow after obtaining information from notification during e2e test
- Fix e2e test for credit card 3D native flow as the web component now contains nested iFrame which makes current test case invalidated.


praveenkumarct:
@leungkinghin Could you please update this PR with discussed changes already?

extension/test/e2e/klarna.spec.js
```diff
@@ -115,18 +131,53 @@ describe('::klarnaPayment::', () => {
...
+        const captureNotification =
+          captureNotificationInteraction.fields.notification
+        if (captureNotification) {
+          const captureNoticationRequestItem =
+            JSON.parse(captureNotification).NotificationRequestItem
+          captureEventCode = captureNoticationRequestItem.eventCode
+          capturePspReference = captureNoticationRequestItem.pspReference
+          captureOriginalPspReference =
+            captureNoticationRequestItem.originalReference
+          captureSuccess = captureNoticationRequestItem.success
```
praveenkumarct, line 166: Good to see that you are already asserting the response of Notification.

praveenkumarct (approved):

extension/test/e2e/klarna.spec.js
leungkinghin: As discussed, it is still good to implement as long as notification works. Otherwise we check the issue of notification in next PR.

PR #1075, opened 2023-02-14
V5 migration get payment methods

leungkinghin:
- Restore getPaymentMethods handler
- Restore unit test and integration test for getPaymentMethods handler, based on the current test case in master branch


extension/resources/web-components-payment-type.json
```diff
@@ -27,6 +27,28 @@
...
+        "en": "getPaymentMethodsResponse"
+      },
+      "type": {
+        "name": "String"
+      },
+      "inputHint": "MultiLine",
+      "required": false
+    },
     {
       "name": "makePaymentRequest",
```
praveenkumarct, line 53: Do we still need this?

praveenkumarct (approved):

extension/resources/web-components-payment-type.json
leungkinghin: Yes. Otherwise Adyen API request and response cannot be put into payment custom fields.

By the way, I removed obsoleted fields from this JSON
https://github.com/commercetools/commercetools-adyen-integration/pull/1075/commits/84dbf1e1bdb153ed349ce6cc2864b777dd223e74

PR #1076, opened 2023-02-15
V5 migration doc

leungkinghin:


praveenkumarct (approved):
LGTM !!
Except those old ADR's which still explains about the old methods.
For ex: https://github.com/commercetools/commercetools-adyen-integration/blob/v5-migration-doc/docs/adr/0005-adding-line-items.md#context

leungkinghin:
> LGTM !! Except those old ADR's which still explains about the old methods. For ex: https://github.com/commercetools/commercetools-adyen-integration/blob/v5-migration-doc/docs/adr/0005-adding-line-items.md#context

ADR can be considered as logs of architectural change. It also helps the future contributors to understand our previous decisions even they are obsoleted. I would rather keep old ADRs as long as we have already provided ADR for web component v5 migration.
https://github.com/commercetools/commercetools-adyen-integration/blob/v5-migration-doc/docs/adr/0012-remove-obsoleted-api-call.md

praveenkumarct (approved):

PR #1081, opened 2023-02-26
Migrate refund payment handler integration test to e2e

praveenkumarct:
TODO: 
Refund payment test fails because, Adyen responding with refund failure status but not sure what is the reason.
I can't see the reason field in the response(May be we should check with Adyen support).


leungkinghin:
@praveenkumarct I checked that refund can only be done if payment is captured. For this reason i move your implementation to Klarna payment since we have implemented payment capture there.

PR #1100, opened 2023-07-18
Cloudformation doesnt support a handler containing a period

samriley:
Cloudformation doesnt support a period in the handler, rename it to an underscore to work around this limitation


samriley:
This will be a breaking change so not sure how open you'll be to accepting this as is. If you have any suggestions to make this change acceptable let me know.

merijndebeer-adyen:
Thank you for your contribution. We have reviewed this pull request and decided to not merge it since it does not fit in the current version of the integration. Regards Adyen Product Management

PR #1103, opened 2023-07-26
Fix certification failure for Commercetools Connect

leungkinghin:
- Move c8 to `devDependencies` in `package.json` as it uses vulnerable dependencies
- Exclude integration test and e2e test from `npm run test` command for Commercetools Connect, and execute `npm run test-ci` in Github Action
- Use mock config in notification module unit test
- Update documentation


praveenkumarct (commented):
You don't need to run Integration and e2e tests completely in the main branch?

I think still there are other users using this repo.
May you could consider ignoring only those tests which are failing at the moment. We can fix them later(Try to create a issue for our reference).

praveenkumarct (approved):

praveenkumarct:
> You don't need to run Integration and e2e tests completely in the main branch?
> 
> I think still there are other users using this repo. May you could consider ignoring only those tests which are failing at the moment. We can fix them later(Try to create a issue for our reference).

I misunderstood the changes, Yes the code looks good for me.

leungkinghin:
> You don't need to run Integration and e2e tests completely in the main branch?
> 
> I think still there are other users using this repo. May you could consider ignoring only those tests which are failing at the moment. We can fix them later(Try to create a issue for our reference).

The change in this PR would not affect the integration test and e2e test in CI/CD. Since Commercetools Connect does not get the adyen config from env-var during build, that's why we add more npm commands (`npm run test` is now dedicated for Commercetools Connect).

PR #1183, opened 2024-07-11
feat: more logs for extension and notification apps

Krysik:
## Summary

The motivation behind this PR is that I would like to observe better what's going on in the extension and notification applications. I would like to track a whole request path by request id and have a response time.

In my PR I inject a logger instead of importing it from utils. The injected logger has a context of the request, it contains information about the requesting path, used HTTP method, and request id. In my opinion, it's better to have a context instead of a raw logger without anything.

Also, I changed the notification handler to accept an object. This approach is more flexible in case of adding new things in the future and it provides more context for the developer, especially for the boolean values, now you don't have to remember the order of the arguments. 

## Tested scenarios

Not much to test here, I added logs 😉😄

### DIFF
--- a/notification/src/handler/notification/notification.handler.js
+++ b/notification/src/handler/notification/notification.handler.js
@@ -5,23 +5,21 @@ import { validateHmacSignature } from '../../utils/hmacValidator.js'
 import utils from '../../utils/commons.js'
 import ctp from '../../utils/ctp.js'
 import config from '../../config/config.js'
-import { getLogger } from '../../utils/logger.js'
 
-const mainLogger = getLogger()
-
-async function processNotification(
+async function processNotification({
   notification,
   enableHmacSignature,
   ctpProjectConfig,
-) {
-  const logger = mainLogger.child({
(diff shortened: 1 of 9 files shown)


Krysik:
@jhilden @lojzatran @emmenko, Will some of you have some time to look at it?

logeecom:
Hello @Krysik 

Our team will look into this and get back to you.

Best regards.

Krysik:
@logeecom any update on this PR?

notification/src/handler/notification/notification.handler.js
```diff
@@ -5,23 +5,21 @@ import { validateHmacSignature } from '../../utils/hmacValidator.js'
...
-const mainLogger = getLogger()
-
-async function processNotification(
+async function processNotification({
   notification,
   enableHmacSignature,
   ctpProjectConfig,
-) {
-  const logger = mainLogger.child({
+  logger
```
vasily-polonsky, line 13: breaking change

PR #1249, opened 2025-04-11
fix(lambda): Set proper handler return value for lambda as ALB target

ethann:
Lambda in AWS can be configured as Target Group for ALB. API Gateway, Lambda Function URL and Application Load Balancer expect lambda to return an object with specific fields, e.g. `statusCode` which have a special behavior (response object). Not specifying `statusCode` field in lambda and triggering it by API Gateway or Function URL converts return object into a string and returns it with status code 200. ALB throws 502 Bad Gateway error instead, expecting lambda to return a proper response object structure with at least `statusCode` field (the rest including `body` seems to be optional).

To keep it compatible with all three solutions and avoid potential conflict between returned body and response object structure, object returned by the lambda should have a proper structure. To keep it compatible with current behavior, status is set to 200 and the body is stringified and returned as the lambda response object `body` field.

API Gateway response body:
https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html#apigateway-types-transforms

Lambda Function url response body (follows API Gateway structure):
https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-payloads

ALB Target group response body:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#respond-to-load-balancer

### DIFF
--- a/extension/index.lambda.js
+++ b/extension/index.lambda.js
@@ -9,8 +9,8 @@ let handler = async (event) => {
   try {
     const body = event.body ? JSON.parse(event.body) : event
     paymentObj = body?.resource?.obj
-    if (!paymentObj)
-      return {
+    if (!paymentObj) {
+      const responseBody = {
         responseType: 'FailedValidation',
         errors: [
           {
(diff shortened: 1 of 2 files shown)


adyen-integrations-support:
Hi @ethann 

The PR has been merged. The unit tests were failing with the proposed fixes, so we addressed the issues and got them passing.

The changes will be available in the next release.

Kind regards

PR #1306, opened 2026-05-09
Fix shopperIP

MarioNaise:
## Summary
Fix `shopperIP` field in `makePaymentRequest` and `createSessionRequest` objects:
Currently, `payment.controller.js` unconditionally sets the `shopperIP` field on `paymentObject` to the value from the requests `x-forwarded-for` header and then applies it to the initial request object later on.
This means that any `shopperIP` provided by the initial request gets lost.
Additionally, the `x-forwarded-for` header contains a wrong value (most likely the Commercetools server IP), so it should not be used as a fallback at all.
Instead, if `shopperIP` is missing in the clients request, it should just stay `undefined`.

## Tested scenarios
- `makePaymentRequest` contains `shopperIP` -> `shopperIP` will be forwarded to Adyen
- `makePaymentRequest` doesn't contain `shopperIP` ->`shopperIP` stays `undefined`
- `createSessionRequest` contains `shopperIP` -> `shopperIP` will be forwarded to Adyen
- `createSessionRequest` doesn't contain `shopperIP` -> `shopperIP` stays `undefined`

**Fixed issue**:

### DIFF
--- a/extension/src/api/payment/payment.controller.js
+++ b/extension/src/api/payment/payment.controller.js
@@ -28,7 +28,6 @@ async function processRequest(request, response, logger) {
   try {
     const authToken = getAuthorizationRequestHeader(request)
     paymentObject = await _getPaymentObject(request, logger)
-    paymentObject.shopperIP = getIPAddressRequestHeader(request)
     const paymentLogger = logger.child({ paymentId: paymentObject.id })
     paymentLogger.debug(
       'Received payment object',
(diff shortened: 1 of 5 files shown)


andihalberkamp:
Hi @merijndebeer-adyen ,

A colleague has provided a fix for the issue described above. This is an important issue for us because it prevents the correct IP address from being sent to Adyen.

As a result, the 3D Secure popup is shown to customers more frequently than necessary, which could negatively impact our conversion rate.

It would be great if you could review the PR when you have a moment. If you have any questions, please feel free to contact me or @MarioNaise.

merijndebeer-adyen:
Thank you @andihalberkamp and @MarioNaise. I will merge the fix into the main branch.

